PR #16, opened 2025-06-14
Add ruby support

### LINKED ISSUE #10: Add ruby instrumentation package
Add the ruby instrumentation to the supported instrumentations. See https://github.com/open-telemetry/opentelemetry-operator/pull/3756 for prior art.

atoulme:
Fixes #10


atoulme:
This PR doesn't run the ruby integration test, which fails. If #19 is merged, we can rebase off it and add ruby to run as integration test, and it will be obvious there is a problem.

The issue seems to come from how ruby autoinstrumentation itself is triggered. Because I took a minimalist approach and ran this with a simple ruby file, there is no bundler in the require path, and therefore it looks like the app doesn't load up the opentelemetry dependencies. The fix is to test with something a bit more realistic, such as a rails app. I asked the ruby devs and they told me they use `rails new my_app --skip-test` as a way to test. That will be the next step.

PR #45, opened 2025-07-20
[chore] add release script

### LINKED ISSUE #7: Create a release workflow
We should have a release workflow that picks up the deb/rpm and other artifacts and publishes them on a github release when running on a tag starting with `v`.

atoulme:
Fixes #7

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,25 @@
+name: Automation - Prepare Release
+
+on:
+  workflow_dispatch:
+    # Determine the version number that will be assigned to the release.
+    inputs:
+      candidate:
+        required: true
+        description: Release candidate version (like 0.70.0)
+jobs:
+  # Releasing opentelemetry-injector
+  prepare-release:
+    runs-on: ubuntu-24.04
+    steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+*"
```
mmanciop: I am not sure what is the value of semver for something that is effectively an amalgamation of a bunch of upstream packages with separate semvers.

It seems unrealistic to me that we'd be able to have any meaningful semver semantics in the versioning of this repo.

I'd just make an index starting from 1.

atoulme: we can run on any tag if you prefer. We might like to discuss the version scheme - ok I'll open an issue for that.

atoulme: We discussed in SIG meeting on 8/18 and will update to run on any tag. We will likely want to test this with a simple release based on the existing libc code before we merge in the zig code.

basti1302: > We will likely want to test this with a simple release based on the existing libc code before we merge in the zig code.

Sounds like that ship has sailed now :-)

mmanciop (approved):

PR #63, opened 2025-08-18
chore: import Dash0 Zig injector

basti1302:
This replaces the current C-based injector code (originally donated by Splunk) with the injector written in Zig, donated by Dash0. The original code this has been sourced from is
https://github.com/dash0hq/dash0-operator/tree/9beb2a95adf3b0bba3e2f75ce7cb6a4c0ff5caac/images/instrumentation/injector.

**Status: Ready for Review**

# Comparison of the existing injector with the new injector

Since this is basically a complete rewrite, it is worth looking at pros and cons for this move:
* The current injector binds a libc — how does that behave with muslc based systems? Can the current binary be used in containers where the libc flavor is not known ahead of times? The new injector does not bind a libc and can be used on glibc as well as musl systems. (Which is quite important if you want to add the injector to unknown containers, which is a common use case in Kubernetes.) 
* AFAICT the current injector unconditionally overwrites all env vars via `setenv` with the values listed in one of the conf files. If I am reading that right, this could lead to losing OTEL_RESOURCE_ATTRIBUTES key-value pairs that are set directly on the process? Also NODE_OPTIONS will be overwritten unconditionally with -r /usr/lib/opentelemetry/otel-js/node_modules/@opentelemetry-js/otel/instrument and remove all other NODE_OPTIONS that might have been set otherwise? The new injector does that conditionally. That is, if OTEL_RESOURCE_ATTRIBUTES is already set in the original environment, it will append to it. Existing keys will not be overwritten. Same for NODE_OPTIONS etc.
* The new injector has support for adding Kubernetes resource attributes from within the process, in a fine grained matter, conditionally. You can  have an observability tool (which adds the injector via ld_preload) provide good defaults for K8s-related resource attributes and also service name and friends, but if the process in question already sets a (better) service name etc., the injector will not override it.
* The current injector can inject arbitrary environment variables via the conf files (but then this is limited by ALLOWED_ENV_VARS). The new injector does not have that capability, instead, the env vars that are added/modified are hard coded. I think this is ultimately a distinction without a difference.
* The new injector is not able to override OTEL_RESOURCE_ATTRIBUTES or OTEL_SERVICE_NAME for .NET applications:
    * Reason: The CLR bootstrapping code (implemented in C++) uses `getenv`, so this part works. But when doing `Environment.GetEnvironmentVariable` from within a .NET application, it will bypass the `getenv` override, because the implementation of `Environment.GetEnvironmentVariable` reads `__environ` into a dictionary at startup and then uses that dictionary for all lookups. That is, while the new injector can inject the OTel auto instrumentation and activate tracing for a .NET application, it is not capable of overriding environment variables that the .NET OTel SDK looks up from within the CLR via `Environment.GetEnvironmentVariable`. 
    * Other runtimes (Node.js, JVM) are not affected by this issue.
    * The `setenv` based approach of the current injector probably does not have that issue (to be verified).

# Questions

There are some open questions for discussion:
- [ ] The arm64 packages for the current injector do not instrument .NET. due to https://github.com/open-telemetry/opentelemetry-injector/blob/d41f25f73143adbf0a3041aa467afcee31c851fc/packaging/fpm/common.sh#L110-L113. Do we want to keep it that way or do we want to enable .NET instrumenation for arm64 as a side effect of landing this? Also, what was the specific reason for disabling it for arm64 so far. 
- [ ] The new injector does not check the executable name, like the currently existingt injector does in https://github.com/open-telemetry/opentelemetry-injector/blob/main/src/main.c#L27-L36. Should we bring this check back? I think this is a two-sided sword:
    - Other executables will not randomly look up NODE_OPTIONS etc., so this should not be needed
    - Some environments will use other executable names (like npm start could have npm as an executable but in fact it is also Node.js, or very old Ubuntu distros (IIRC) used nodejs instead of node (https://askubuntu.com/a/405035)? Of course we can work around the known cases with a growing list of executable names.

# Possible follow up tasks

- [ ] Look into the `systemd` specific stuff, that is `packaging/fpm/examples/systemd/00-opentelemetry-injector.conf` and the relevant section in README.md. I did not update those. It is a bit unclear to me how that relates to the injector (I think it does not relate to the injector at all and just uses the rpm/debian package as a vehicle that bundles all of the auto-instrumentation agents?)
- [ ] Why do we upload the rpm/deb in the build-package, independent of the packaging-integration-tests, and where do we upload them to? For what purpose?
- [ ] The .NET auto-instrumentation agent depends on the CPU architecture, make sure the packaging scripts for building the RPM/DEB include the correct version. See also: https://github.com/dash0hq/dash0-operator/blob/main/images/instrumentation/dotnet/download-instrumentation.sh
- [ ] create a container image distribution: In addition to deb and rpm archives, we should also publish a multi-platform container image for arm64 and amd64 that contains the injector binary, and the auto-instrumentation agents for supported runtimes (.NET, JVM, Node.js at the moment), similar to https://github.com/dash0hq/dash0-operator/tree/main/images/instrumentation / https://github.com/dash0hq/dash0-operator/blob/main/images/instrumentation/Dockerfile.
- [ ] create more comprehensive packaging tests: There are currently tests one single the combination deb & Java. It might be nice to have tests for rpm as well, maybe also for the other runtimes.
- [ ] update to the latest Zig version (15.1 is already out)
- [ ] Add renovate config for Zig
- [ ] Update versions in packaging/java-agent-release.txt etc. for all runtimes to latest. There is already a renovate config for them. I don't see any renovate PRs for those versions though. What gives?
- [ ] Update packaging/fpm/examples/systemd/00-opentelemetry-injector.conf, it is probably no longer accurate.
- [ ] Run packaging integration tests (make integration-test-deb-java etc.) for both architectures on GH actions. Currently they are only run for amd64.
- [ ] Activate shellcheck in `make lint` and align all shell scripts with shellcheck rules.
- [ ] Actually publish files via github releases (injector binaries, rpms, debs)

----

# Open TODOs for this PR (all done now)

There are a couple of things that need to be taken care of before this should be merged:

- [x] add integration tests for .NET.
- [x] revisit the configuration mechanism from the original injector, i.e. the [conf](https://github.com/open-telemetry/opentelemetry-injector/tree/main/packaging/fpm/etc/opentelemetry/otelinject). We need to rewrite some kind of configuration mechanism in Zig. Instead of directly configuring env vars, it might be more desirable to only configure the various locations of auto-instrumentation agents for different runtimes.
- [x] verify that the existing packaging (RPM, DEB etc.) still works
- [x] create CI job for injector integration tests (test/scripts/test-all.sh)
- [x] fix Makefile
- [x] fix CI build
- [x] add linting CI jobs
- [x] update CONTRIBUTING.md
- [x] merge READMEs


README.md
```diff
@@ -32,37 +54,37 @@ The following methods are supported to manually activate and configure Auto Inst
...
+  - `/etc/opentelemetry/otelinject/dotnet.conf`:
+    ```
+    CORECLR_ENABLE_PROFILING=1
+    CORECLR_PROFILER={918728DD-259F-4A6A-AC2B-B85E1B658318}
+    CORECLR_PROFILER_PATH=/usr/lib/opentelemetry/otel-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so
+    DOTNET_ADDITIONAL_DEPS=/usr/lib/opentelemetry/otel-dotnet/AdditionalDeps
+    DOTNET_SHARED_STORE=/usr/lib/opentelemetry/otel-dotnet/store
+    DOTNET_STARTUP_HOOKS=/usr/lib/opentelemetry/otel-dotnet/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll
+    OTEL_DOTNET_AUTO_HOME=/usr/lib/opentelemetry/otel-dotnet
+    ```
```
atoulme: let's land this type of lint change separately

README.md
```diff
@@ -32,37 +54,37 @@ The following methods are supported to manually activate and configure Auto Inst
...
-   - `OTEL_LOGS_EXPORTER`
-   - `OTEL_METRICS_EXPORTER`
-   - `OTEL_RESOURCE_ATTRIBUTES`
-   - `OTEL_SERVICE_NAME`
+  - `OTEL_EXPORTER_OTLP_ENDPOINT`
+  - `OTEL_EXPORTER_OTLP_PROTOCOL`
+  - `OTEL_LOGS_EXPORTER`
+  - `OTEL_METRICS_EXPORTER`
+  - `OTEL_RESOURCE_ATTRIBUTES`
+  - `OTEL_SERVICE_NAME`
```
atoulme: same

README.md
```diff
@@ -32,37 +54,37 @@ The following methods are supported to manually activate and configure Auto Inst
...
 
    Check the following for details about these environment variables and default values:
-   - [Java](https://opentelemetry.io/docs/zero-code/java/agent/configuration/)
-   - [Node.js](https://opentelemetry.io/docs/zero-code/js/configuration/)
-   - [.NET](https://opentelemetry.io/docs/zero-code/dotnet/configuration/)
+  - [Java](https://opentelemetry.io/docs/zero-code/java/agent/configuration/)
+  - [Node.js](https://opentelemetry.io/docs/zero-code/js/configuration/)
+  - [.NET](https://opentelemetry.io/docs/zero-code/dotnet/configuration/)
```
atoulme: same

README.md
```diff
@@ -85,24 +107,24 @@ The following methods are supported to manually activate and configure Auto Inst
...
+  - .NET
+    ```
+    DefaultEnvironment="CORECLR_ENABLE_PROFILING=1"
+    DefaultEnvironment="CORECLR_PROFILER={918728DD-259F-4A6A-AC2B-B85E1B658318}"
+    DefaultEnvironment="CORECLR_PROFILER_PATH=/usr/lib/opentelemetry/otel-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so"
+    DefaultEnvironment="DOTNET_ADDITIONAL_DEPS=/usr/lib/opentelemetry/otel-dotnet/AdditionalDeps"
+    DefaultEnvironment="DOTNET_SHARED_STORE=/usr/lib/opentelemetry/otel-dotnet/store"
+    DefaultEnvironment="DOTNET_STARTUP_HOOKS=/usr/lib/opentelemetry/otel-dotnet/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll"
+    DefaultEnvironment="OTEL_DOTNET_AUTO_HOME=/usr/lib/opentelemetry/otel-dotnet"
+    ```
```
atoulme: it appears these are whitespace changes only - ok to push them separately

README.md
```diff
@@ -114,9 +136,9 @@ The following methods are supported to manually activate and configure Auto Inst
    EOH
    ```
    Check the following for all supported environment variables and default values:
-   - [Java](https://opentelemetry.io/docs/zero-code/java/agent/configuration/)
-   - [Node.js](https://opentelemetry.io/docs/zero-code/js/configuration/)
-   - [.NET](https://opentelemetry.io/docs/zero-code/dotnet/configuration/)
+  - [Java](https://opentelemetry.io/docs/zero-code/java/agent/configuration/)
+  - [Node.js](https://opentelemetry.io/docs/zero-code/js/configuration/)
+  - [.NET](https://opentelemetry.io/docs/zero-code/dotnet/configuration/)
```
atoulme: same

test/bin/.gitignore
```diff
@@ -0,0 +1 @@
+*.so
```
atoulme: best to leave that in one file at the root of the repo?

test/scripts/util
```diff
@@ -0,0 +1,4 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+
```
atoulme: I guess this file can be safely removed

zig-build.sh
```diff
@@ -0,0 +1,24 @@
...
+# See also: ./watch-zig-build.sh
+# The point is that `zig build` does not print anything when there have been no changes since the last successful build,
+# which makes it hard to tell (when using entr) whether the build is still ongoing or there simply was nothing to do.
+# This script will simply print a message indicating the successful build to circumvent this.
+
+set -euo pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+if zig build --prominent-compile-errors --summary none; then
```
atoulme: can this be a make target?

zig-test.sh
```diff
@@ -0,0 +1,24 @@
...
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# A small helper script intended to be used with entr, like so:
+# fd | entr ./zig-test.sh
+# ...to get fast feedback on test errors when working on the Zig code.
+# See also: ./watch-zig-test.sh
+# The point is that `zig build test` does not print anything when there have been no changes since the last successful
+# build, which makes it hard to tell (when using entr) whether the build is still ongoing or there simply was nothing
+# to do. This script will simply print a message indicating the successful build to circumvent this.
```
atoulme: that's good context

zig-test.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -euo pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+if zig build test --prominent-compile-errors --summary none; then
```
atoulme: same can this be a make target?

zig-version
```diff
@@ -0,0 +1,3 @@
+# See https://dl-cdn.alpinelinux.org/alpine/edge/community, check for zig-*.apk for available versions.
+# Keep this in sync with .github/workflows/ci.yaml -> mlugg/setup-zig.with.version.
+ZIG_VERSION=0.14.1-r0
```
atoulme, line 3: will this be updated by renovate?

Makefile
```diff
@@ -1,4 +1,4 @@
-ARCH=amd64
+ARCH?=amd64
```
atoulme: can split that into a separate PR

README.md
basti1302: I'll revert this. This was not deliberately changed, probably some auto-format of my IDE.

Makefile
basti1302: reverted

test/bin/.gitignore
basti1302: good catch, the directory is actually no longer used, the tests build and use the binaries from /dist now. I'll remove this.

zig-build.sh
basti1302: all four scripts (zig-build.sh, zig-test.sh, watch-zig-build.sh and watch-zig-test.sh) are now make targets, the scripts have been removed.

zig-version
basti1302: No, auto-update for this is not in place yet.

README.md
```diff
@@ -1,9 +1,27 @@
...
+[`/etc/ld.so.preload`](https://man7.org/linux/man-pages/man8/ld.so.8.html#FILES) file, or similar mechanisms to inject
+environment variables into processes at startup.
+
+It serves two main purposes:
+* Inject an OpenTelemetry Auto Instrumentation agent into the process to capture and report distributed traces and
+  metrics to the [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/) for supported runtimes.
+* Set resource attributes automatically, (for example Kubernetes related resource attributes and service related
+  resource attributes in environments where this is applicable).
+
+The injector can be used to enable automatic zero-touch instrumenation of processes.
```
atoulme:
```suggestion
The injector can be used to enable automatic zero-touch instrumentation of processes.
```

README.md
```diff
@@ -22,51 +40,88 @@ The following methods are supported to manually activate and configure Auto Inst
...
+   ```
+   dotnet_auto_instrumentation_agent_path_prefix=/usr/lib/opentelemetry/dotnet
+   jvm_auto_instrumentation_agent_path=/usr/lib/opentelemetry/javaagent.jar
+   nodejs_auto_instrumentation_agent_path=/usr/lib/opentelemetry/otel-js/node_modules/@opentelemetry-js/otel/instrument
+   ```
+   There is usually no need to modify this file, unless you want to provide your own instrumentation files.
+
+   However, the configuration file `/etc/opentelemetry/otelinject.conf` can also be used to selectively disable
+   auto-instrumentation for a specific runtime, by setting the respective path to an empty string.
+   For exapmle, the following file would leave JVM and Node.js auto-instrumentation active, while disabling .NET
```
atoulme:
```suggestion
   For example, the following file would leave JVM and Node.js auto-instrumentation active, while disabling .NET
```

README.md
```diff
@@ -22,51 +40,88 @@ The following methods are supported to manually activate and configure Auto Inst
...
+   These aforementioned environment variables can also be used to selectively disable auto-instrumentation for a
+   specific runtime, by setting the respective variable to an empty string, that is, set:
+    - `DOTNET_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX=""` to disable .NET auto-instrumentation
+    - `JVM_AUTO_INSTRUMENTATION_AGENT_PATH=""` to disable JVM auto-instrumentation
+    - `NODEJS_AUTO_INSTRUMENTATION_AGENT_PATH=""` to disable Node.js auto-instrumentation
 3. Reboot the system or restart the applications/services for any changes to take effect. The `libotelinject.so` shared
    object library will then be preloaded for all subsequent processes and inject the environment variables from the
    `/etc/opentelemetry/otelinject` configuration files for Java and Node.js processes.
 
+When providing you own instrumentation files (for example via environment variables like `DOTNET_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX`) the following directory structure is expected:
```
atoulme:
```suggestion
When providing your own instrumentation files (for example via environment variables like `DOTNET_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX`) the following directory structure is expected:
```

README.md
```diff
@@ -129,5 +184,6 @@ See [CONTRIBUTING.md](CONTRIBUTING.md).
 - [Antoine Toulme](https://github.com/atoulme), Splunk
 - [Jacob Aronoff](https://github.com/jaronoff97), Omlet
 - [Michele Mancioppi](https://github.com/mmanciop), Dash0
+- [Bastian Krol](https://github.com/mmanciop), Dash0
```
atoulme:
```suggestion
- [Bastian Krol](https://github.com/basti1302), Dash0
```

test/scripts/create-no-sdk-dummy-files.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+echo "not creating and OTel auto instrumentation dummy files"
```
atoulme:
```suggestion
echo "not creating any OTel auto instrumentation dummy files"
```
not sure

atoulme (approved):

README.md
```diff
@@ -22,51 +40,88 @@ The following methods are supported to manually activate and configure Auto Inst
...
-   - [.NET](https://opentelemetry.io/docs/zero-code/dotnet/configuration/)
+   Alternatively, set the environment variable `LD_PRELOAD=/usr/lib/opentelemetry/libotelinject.so` for a specific
+   process to activate auto-instrumentation for tha process. For example:
+   ```
+   LD_PRELOAD=/usr/lib/opentelemetry/libotelinject.so node myapp.js
+   ```
+2. The default configuration file `/etc/opentelemetry/otelinject.conf` includes the required settings, i.e. the paths to
+   the respective auto-instrumentation agents per runtime:
+   ```
+   dotnet_auto_instrumentation_agent_path_prefix=/usr/lib/opentelemetry/dotnet
```
jaronoff97, line 59: can be done in a follow up, but would be great to have two tables – one for all the available flags for the .conf file and one for the available environment variables.

jaronoff97: i guess this is one table because of the code structure.

src/config.zig
```diff
@@ -0,0 +1,453 @@
...
+        const line = line_buffer_array.toOwnedSlice() catch |e| {
+            print.printError("error in toOwnedSlice while reading configuration file {s}: {}", .{ cfg_file_path, e });
+            break;
+        };
+        defer alloc.page_allocator.free(line);
+        _ = parseLine(line, cfg_file_path, configuration);
+    }
+}
+
+test "readConfigurationFile: file does not exist" {
```
jaronoff97, line 145: I'm still getting used to the zig syntax and structure, is it common to include tests in the same file as the code?

src/config.zig
```diff
@@ -0,0 +1,453 @@
...
+            '\n',
+            max_line_length,
+        ) catch |err| switch (err) {
+            error.EndOfStream => {
+                const line = line_buffer_array.toOwnedSlice() catch |e| {
+                    print.printError("error in toOwnedSlice while reading configuration file {s}: {}", .{ cfg_file_path, e });
+                    break;
+                };
+                defer alloc.page_allocator.free(line);
+                _ = parseLine(line, cfg_file_path, configuration);
```
jaronoff97, line 112: I would think the end of stream implies there's no lines left to parse, can we not just break here?

src/jvm.zig
```diff
@@ -0,0 +1,419 @@
...
+    // behavior, we refrain from merging OTEL_RESOURCE_ATTRIBUTES in that scenario.
+    if (original_java_tool_options_env_var_value_optional) |original_java_tool_options_env_var_value| {
+        // If JAVA_TOOL_OPTIONS is already set, append our values.
+        if (new_resource_attributes_optional) |new_resource_attributes| {
+            defer alloc.page_allocator.free(new_resource_attributes);
+
+            if (std.mem.indexOf(u8, original_java_tool_options_env_var_value, "-Dotel.resource.attributes=")) |startIdx| {
+                // JAVA_TOOL_OPTIONS already contains -Dotel.resource.attribute, we need to merge the existing with the new key-value list.
+                const actualStartIdx = startIdx + 27;
+                // assume the list of key-value pairs is terminated by a space of the end of the string
```
jaronoff97: I think it's also possible for it to be terminated by another quote quote as well i.e. -Dotel.resource.attributes="key.value=hello.world"

src/root.zig
```diff
@@ -0,0 +1,147 @@
...
+
+export fn getenv(name_z: types.NullTerminatedString) ?types.NullTerminatedString {
+    const name = std.mem.sliceTo(name_z, 0);
+
+    // Need to change type from `const` to be able to lock
+    var env_mutex = _env_mutex;
+    env_mutex.lock();
+    defer env_mutex.unlock();
+
+    // Dynamic libs do not get the std.os.environ initialized, see https://github.com/ziglang/zig/issues/4524, so we
```
jaronoff97, line 44: im slightly confused about why we need to re-initialize this with a lock on every call? I understand we cannot rely on it being present on start, but post start shouldn't it continue to be present?

jaronoff97 (approved):
Nothing blocking, just a few questions about things I couldnt fully grasp. Everything else makes sense here. Thank you!

src/config.zig
jaronoff97: it seems like sometimes we do this, other times we don't

README.md
basti1302: Yes, let's improve that in a follow up. 👍

src/config.zig
basti1302: Unfortunately not. `streamUntilDelimiter` gives you the very last line (which is not terminated by `\n`) together with the error.EndOfStream. So if we omit the parseLine call here, we would miss the very last line of the file. But I see where you are coming from, I'll add a comment here to explain that.

src/config.zig
basti1302: Yes, I think this is a common idiom. See https://ziglang.org/documentation/0.15.1/#toc-Test-Declarations:

> Test declarations can be written in the same file, where code under test is written, or in a separate Zig source file. Since test declarations are top-level declarations, they are order-independent and can be written before or after the code under test.

Also: https://www.reddit.com/r/Zig/comments/wl6rru/how_do_you_organize_your_tests/

There is one key difference: tests in the same file have access to _all_ functions, tests in a separate file only have access to `pub` functions. I think I mainly stuck with the "test in the same file" approach, except for more invasive tests that set and remove environment variables (print_test, resource_attributes_test). There is a comment that explains this at the top of both of these *_test.zig files. Granted there is probably no real need to have the latter in separate files, I don't have a strong opinion about that.

src/jvm.zig
basti1302: Oh, very good find. I'll fix this.

src/root.zig
basti1302: I am not sure I follow. What do you mean specifically with "re-initialize"? As far as I can tell, the lock struct is only initialized once, in line 25.

Or is your question whether we actually need to _lock_ every time `getenv` is called? That I am not 100% sure about, but more importantly, I am not sure that we would *not* run into weird race conditions for parallel calls to getenv without the mutex.

README.md
basti1302: Oh boy... 🤦

src/root.zig
jaronoff97: kind of – my question is less why we need the lock (it's mean to be locking the setting of std.os.environ), and rather why we continue to rebind the value of std.os.environ. Maybe put differently, it's my understanding of the code that each time getenv is called we rebind std.os.environ, not each invocation of the injector code, but rather each internal call of getenv is rebinding. Is my understanding correct? (we're pushing the limits of my zig knowledge here, so apologies for any ignorance)

basti1302: Ah, gotcha. Yes, your understanding is correct, we re-initialize std.os.environ from __environ on every `getenv` call. We could probably switch to an approach where we only do that on the first `getenv` call.

basti1302: I created an issue for this: https://github.com/open-telemetry/opentelemetry-injector/issues/67

PR #79, opened 2025-09-11
[chore] add ARCH to zig build targets

jackshirazi:
Otherwise those targets build for the default arc64 arch rather than $ARCH


basti1302 (approved):

basti1302:
I have created https://github.com/open-telemetry/opentelemetry-injector/pull/85 to deal with the Tomcat download issue.

PR #113, opened 2025-10-31
[chore] add git to build to suppress errors

atoulme:
the build complains about git missing and the code not being in a git repository. Mounting the git folder and making git available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN apk add --no-cache make
 
 COPY zig-version /otel-injector-test-build/zig-version
 RUN source /otel-injector-test-build/zig-version && \
-  apk add zig="$ZIG_VERSION" --repository=https://dl-cdn.alpinelinux.org/alpine/edge/community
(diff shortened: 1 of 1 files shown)


basti1302 (changes requested):
I came across [this](https://github.com/open-telemetry/opentelemetry-injector/blob/2cad62532c8e01eddb6b1a38c5448fc2c28cc4c6/Makefile#L132) as well when working on the injector build. To be honest, I'm not a fan of solving it this way. IIRC, the container image build does not really need git nor access to the git repository.

Adding the `.git` directory makes the image build context bigger without any actual reason.

I would rather untangle the dependency to git in the Makefile. Unfortunately, the dependency to the git executable is not in any specific target but in the top level declaration of `SRC_ROOT`, which is only used for chloggen, right? Maybe we can move all of this into the chloggen targets somehow?

Or figure out the SRC_ROOT without using git, like so: `SRC_ROOT ?= $(shell pwd)`?

atoulme:
That makes sense, I prefer lazy makefile targets too. I will fix it the way you prescribe.

PR #127, opened 2025-11-06
[nodejs] add testing and fix issues

atoulme:
Testing identified issues with the RPM and Debian packages.
The Debian and RPM packages now contain the nodeJS dependencies organized in a folder structure rather than an archive.
The configuration points correctly to the file to load to launch autoinstrumentation.
  
NodeJS tests now run as part of CI.


basti1302 (approved):
👍 👍

PR #129, opened 2025-11-08
[chore] remove git error when building

atoulme:


basti1302 (approved):
Nice!

atoulme:
Looks like changelog fails. I will check

PR #131, opened 2025-11-09
[chore] test both amd64 and arm64 archs

atoulme:
Test both amd64 and arm64 when possible. Tests for .NET with arm64 were treacherous, so I abandoned testing that particular combination.


atoulme:
https://github.com/open-telemetry/admin/pull/409 is required to merge this change.

PR #136, opened 2025-11-10
[chore] activate shellcheck in `lint` target

sultanowskii:
Resolves https://github.com/open-telemetry/opentelemetry-injector/issues/74

### DIFF
--- a/packaging/fpm/common.sh
+++ b/packaging/fpm/common.sh
@@ -72,7 +72,7 @@ download_nodejs_agent() {
     mkdir -p "js"
     pushd "js"
     npm pack "@opentelemetry/auto-instrumentations-node@${tag#v}"
-    mv *.tgz otel-js.tgz
+    mv ./*.tgz otel-js.tgz
     npm install --global=false otel-js.tgz
     rm otel-js.tgz
     popd


packaging/fpm/deb/build.sh
```diff
@@ -5,8 +5,10 @@
 
 set -euxo pipefail
 
-SCRIPT_DIR="$( cd "$( dirname ${BASH_SOURCE[0]} )" && pwd )"
-. $SCRIPT_DIR/../common.sh
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# shellcheck disable=SC1091 # Including common.sh
```
basti1302, line 10: Another option would be to tell shellcheck where to find common.sh, instead if disabling it, i.e. something like

```
# shellcheck source=packaging/fpm/common.sh‎
. "$SCRIPT_DIR/../common.sh"
```

That way, common.sh would also be included in the linting. If we decide to do that, the same change would need to be applied to packaging/fpm/rpm/build.sh.

But we can also take care of that in a follow-up PR, to not increase the scope of this one, that's fine to.

basti1302 (approved):
Awesome! Thanks so much for contributing this.

packaging/fpm/deb/build.sh
basti1302: Actually, nevermind, since common.sh also is found by the pattern, it is inspected by shellcheck anyway, so this is fine as it is.

packaging/fpm/common.sh
```diff
@@ -70,8 +71,8 @@ download_nodejs_agent() {
     pushd "$(dirname "$dest")"
     mkdir -p "js"
     pushd "js"
-    npm pack @opentelemetry/auto-instrumentations-node@${tag#v}
-    mv *.tgz otel-js.tgz
+    npm pack "@opentelemetry/auto-instrumentations-node@${tag#v}"
+    mv ./*.tgz otel-js.tgz
```
basti1302: q: was this mandated by shellcheck? I don't see why adding the `./` prefix would be necessary here.

packaging/fpm/common.sh
```diff
@@ -82,6 +83,7 @@ download_dotnet_agent() {
     local tag="$1"
     local dest="$2"
 
+    # shellcheck disable=SC2153 # ARCH is defined in Makefile
```
basti1302: nit: instead of disabling it, we could use something like `case "${ARCH:-}" in` to fix this.

packaging/fpm/common.sh
sultanowskii: Yes, https://www.shellcheck.net/wiki/SC2035

sultanowskii:
Regarding the issues CI has ran into - I couldn't reproduce them on my machine - `make lint` just doesn't give me these warnings

I've looked into it - looks like in CI, shellcheck is of some old version, while I have 0.11.0

Example (https://www.shellcheck.net/wiki/SC2002), quote:
> In later versions, it is [optional](https://www.shellcheck.net/wiki/optional). It must be explicitly enabled with a [directive](https://www.shellcheck.net/wiki/directive) enable=useless-use-of-cat

So for now I've added `enable-useless-use-of-cat` for local vs CI compliance and added the explicit ignoring of `SC2317`, but it might be a good idea to bump its version (though, I have no clue where is it being installed - it isn't mentioned in .github workflows at all)

packaging/fpm/common.sh
basti1302: Ah, interesting. The more you learn :-)

PR #148, opened 2025-11-26
environment settings support for agents

### LINKED ISSUE #82: Support environment settings for agent
### Component(s)

All

With the operator, I can auto-instrument everything in a namespace, and/or be selective and auto-instrument anything deployed with a specific annotation. Additionally the definition of the auto-instrumention can include additional env, eg OTEL_EXPORTER_OTLP_ENDPOINT. So there is nothing needed to be added apart from one annotation, and the process is auto-instrumented including sending to the collector.

With the injector, I can auto-instrument everything in the system, or per-process, but if I want to have eg OTEL_EXPORTER_OTLP_ENDPOINT defined, I need to add that separately using /etc/environment or user init file or directly per process. This is a lot more steps

Support of environment variables associated with the agent. So for example one way this could be done would be in the `otelinject.conf`, the file could look something like

```
all_auto_instrumentation_agent_env=OTEL_EXPORTER_OTLP_ENDPOINT=https://somewhere
all_auto_instrumentation_agent_env=OTEL_EXPORTER_OTLP_HEADERS="Authorization=ApiKey BLAH"
dotnet_auto_instrumentation_agent_path_prefix=/usr/lib/opentelemetry/dotnet
jvm_auto_instrumentation_agent_path=/usr/lib/opentelemetry/javaagent.jar
jvm_auto_instrumentation_agent_env=OTEL_OPTION1=x
jvm_auto_instrumentation_agent_env=OTEL_OPTION2=y
nodejs_auto_instrumentation_agent_path=/usr/lib/opentelemetry/otel-js/node_modules/@opentelemetry-js/otel/instrument
```

Where any env vars defined by `all_auto_instrumentation_agent_env` would be applied to all agent envs, and the individual agents could additional have multiple `<lang>_auto_instrumentation_agent_env` env var definitions that would apply to that agent

- Applying the environment system-wide is problematic because too wide a scope, means that env vars can leak into other processes that could inadvertently use them.
- Applying at the user level is okay from a technical pov, but means the env vars are disassociated from the agent declaration and this seems like a maintenance problem for users. 
- Doing it per process (probably easiest by putting them into a script and running that before starting the process) is technically okay but then there is little point in having the injector since you could also set the script to construct the required env vars (such as `JAVA_TOOL_OPTIONS`) even for the specific architecture, so if I'm doing that already I may as well not use the injector - which means instead of having a nice auto-instrumentation setup, I have to do stuff to get instrumentation

### Additional context

The operator uses the k8s yaml to specify per-instrumentation env, eg [see this example](https://opentelemetry.io/docs/platforms/kubernetes/operator/automatic/#java-excluding-auto-instrumentation). I don't think we need to go the yaml route because defining env vars is pretty straightforward and can be done as properties rather than needing more structure, but I have nothing against a more structured way, it's the functionality that's needed to make the injector usable as a basic one-install auto-instrumentation capability

### Tip

sultanowskii:
Resolves #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ This method requires `root` privileges.
     - `JVM_AUTO_INSTRUMENTATION_AGENT_PATH=""` to disable JVM auto-instrumentation
     - `NODEJS_AUTO_INSTRUMENTATION_AGENT_PATH=""` to disable Node.js auto-instrumentation
 
-3. (Optional) The default env agent configuration file `/etc/opentelemetry/default_auto_instrumentation_env.conf` is empty (use 
+3. (Optional) The default env agent configuration file `/etc/opentelemetry/default_auto_instrumentation_env.conf` is empty (use
    `all_auto_instrumentation_agents_env_path` option to specify other path). You can add environment variables
    in this file - these variables will be passed to all agents' envs.
 


packaging/fpm/etc/opentelemetry/otelinject.conf
```diff
@@ -1,5 +1,9 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
+all_auto_instrumentation_agent_env_path=/etc/opentelemetry/env.conf
 dotnet_auto_instrumentation_agent_path_prefix=/usr/lib/opentelemetry/dotnet
+dotnet_auto_instrumentation_agent_env_path=/etc/opentelemetry/dotnet/env.conf
```
basti1302: Just as a heads up, the next iteration of the injector will use `setenv` unconditionally to inject environment variables. In particular, it will no longer possess the knowledge whether the host process (the process we are injecting env vars into) is the Node.js runtime, a JVM, .NET or anything else.

With that in mind, I think implementing a mechanism that is supposed to only inject environment variables into specific runtimes (dotnet_auto_instrumentation_agent_env_path vs jvm_auto_instrumentation_agent_env_path) becomes unfeasible.

For context, the plan is to upstream the latest changes from here: https://github.com/dash0hq/dash0-operator/tree/main/images/instrumentation/injector

For the details why we plan to switch to a `setenv` based approach, you can read up on that here: https://github.com/dash0hq/dash0-operator/blob/main/images/instrumentation/injector/README.md#design-constraints

sultanowskii: As we've had discussed in the issue, I removed language-specific options

basti1302:
> Oh, by the way, @basti1302, I have a somewhat off-topic question: Why the format of config is not a yaml/toml/json/...?
> 
> I'm not against the current simple key-value approach, but was just wondering why not just go for yaml/toml/json/... format and take a handy 3rd-party parser instead of doing it by hand?

Good question. I don't have a strong opinion on that, to be honest. I think one motivation was to keep the injector binary as small and free of dependencies as possible. Adding a yaml or toml parser would probably be fine though.

There was also some prior art, even before I added the Zig injector code, there was a simpler injector in this repository, written in C, and it had the same key-value pair config file format https://github.com/open-telemetry/opentelemetry-injector/blob/d41f25f73143adbf0a3041aa467afcee31c851fc/tests/java/otelinject.conf, so I think I just went with that.

sultanowskii:
Hi @basti1302 ! I think I mostly finished this one.

But I've run into a problem: for some reason, new integration test just won't pass for JVM and dotnet (passes for nodejs). After a little bit of debugging, it seems like JVM and dotnet processes just don't call our substituted `getenv()` at all.

To see for yourself, you can put something like:

```zig
print.printMessage("getenv({s}) called", .{name_z});
```

in `getenv()` at **root.zig**.

My suspicion is that dotnet and jvm might load all env variables into memory at the startup, therefore all subsequent "getenv" calls don't actually perform a system call.

Could you please take a look at it? How should we proceed?

sultanowskii:
There is no direct way to "setenv" (`std.posix.setenv` doesn't exist). So I figured the simplest approach here is to modify `__environ` and `std.os.environ` :)

src/config.zig
```diff
@@ -16,20 +16,28 @@ const dotnet_path_key = "dotnet_auto_instrumentation_agent_path_prefix";
 const jvm_path_key = "jvm_auto_instrumentation_agent_path";
 const nodejs_path_key = "nodejs_auto_instrumentation_agent_path";
 
+const all_agent_env_path_key = "all_auto_instrumentation_agent_env_path";
```
basti1302: _nit:_ This could be renamed to `all_auto_instrumentation_agents_env_path`. (As it is related to all agents, it should be plural I think.)

src/config.zig
```diff
@@ -45,57 +53,108 @@ pub fn readConfiguration() InjectorConfiguration {
...
         .dotnet_auto_instrumentation_agent_path_prefix = dotnet_default,
         .jvm_auto_instrumentation_agent_path = jvm_default,
         .nodejs_auto_instrumentation_agent_path = nodejs_default,
+        .all_auto_instrumentation_agent_env_path = all_env_default,
+        .all_auto_instrumentation_agent_env_vars = all_agent_env_vars,
     };
 }
 
 fn readConfigurationFile(cfg_file_path: []const u8, configuration: *InjectorConfiguration) void {
+    const fns = struct {
```
basti1302: _question:_ Do we need the surrounding struct `fns` here? Can we just pass `applyKeyValue` to `parseConfiguration` directly, instead of passing `fns.parseConfiguration`?

_praise:_ Nice work extracting the key-value application (and error handling) out into their own function and then reusing `parseConfiguration` in `readAllAgentEnvFile`! 👍

packaging/fpm/etc/opentelemetry/otelinject.conf
```diff
@@ -1,5 +1,6 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
+all_auto_instrumentation_agent_env_path=/etc/opentelemetry/env.conf
```
basti1302: _thougt:_ I think we should choose a different name here. `/etc/opentelemetry/env.conf` sounds quite generic, as if it configures everything related to Open Telemetry on the system. I would prefer if the file name made it clear that it only applies to auto-instrumenting processes.

Maybe something like `/etc/opentelemetry/default_auto_instrumentation_env.conf`?

src/config.zig
```diff
@@ -45,57 +53,108 @@ pub fn readConfiguration() InjectorConfiguration {
...
-            return InjectorConfiguration{
-                .dotnet_auto_instrumentation_agent_path_prefix = "",
-                .jvm_auto_instrumentation_agent_path = "",
-                .nodejs_auto_instrumentation_agent_path = "",
-            };
-        };
-    const nodejs_default =
-        std.fmt.allocPrint(alloc.page_allocator, "{s}", .{default_nodejs_auto_instrumentation_agent_path}) catch |err| {
+    const all_agent_env_vars = std.StringHashMap([]u8).init(alloc.page_allocator);
+    const errorHandler = struct {
```
basti1302: _question:_ Do we need the surrounding struct `errorHandler` here? Can we just declare the function `printErrorReturnEmpty` here and use the function directly in the code below instead of referencing it as `errorHandler.printErrorReturnEmpty`?

src/root.zig
```diff
@@ -48,10 +50,7 @@ export fn getenv(name_z: types.NullTerminatedString) ?types.NullTerminatedString
     // https://git.musl-libc.org/cgit/musl/tree/src/env/setenv.c) if the backing memory location is outgrown by apps
     // modifying the environment via setenv or putenv.
     const environment_optional: [*:null]?[*:0]u8 = @ptrCast(@alignCast(__environ));
-    var environment_count: usize = 0;
-    if (@intFromPtr(__environ) != 0) { // __environ can be a null pointer, e.g. directly after clearenv()
-        while (environment_optional[environment_count]) |_| : (environment_count += 1) {}
-    }
+    const environment_count = countEnivironmentVariables(environment_optional);
```
basti1302: _nit:_ typo: countEnivironmentVariables -> countEnvironmentVariables

src/config.zig
```diff
@@ -45,57 +53,108 @@ pub fn readConfiguration() InjectorConfiguration {
...
+
+    return parseConfiguration(configuration, env_file, env_file_path, fns.applyKeyValue);
+}
+
+fn parseConfiguration(
+    configuration: *InjectorConfiguration,
+    file: std.fs.File,
+    file_path: []const u8,
+    comptime applyKeyValueToConfig: ConfigApplier,
+) void {
```
basti1302, line 147: _issue (blocking):_ Please use the previously used identifiers here, i.e. `config_file` instead of `file` and `cfg_file_path` instead of `file_path`. I agree that within this new function `parseConfiguration`, prefixing the identifiers with config/cfg is redundant, so your name changes are very reasonable, but: There are other open PRs that also touch this code (for example https://github.com/open-telemetry/opentelemetry-injector/pull/152) and keeping the same identifiers would hopefully reduce the inevitable merge conflicts a bit.

src/config.zig
```diff
@@ -258,174 +337,170 @@ test "readConfigurationFile: does not parse overly long lines" {
     );
 }
 
-/// Parses a single line from the configuration file and updates the configuration accordingly.
-/// Returns true if the line is a valid line (key-value pair, an empty line, comment) and false otherwise.
-fn parseLine(line: []u8, cfg_file_path: []const u8, configuration: *InjectorConfiguration) bool {
+/// Parses a single line from a configuration file.
+/// Returns a key-value pair if the line is valid (key-value pair, an empty line, comment)
+/// and null otherwise.
```
basti1302: _issue:_ I think the comment is not entirely accurate. It returns null for empty lines and comments now, right?


```suggestion
/// Returns a key-value pair if the line is a valid key-value pair, and null for empty
/// lines, comments and invalid lines.
```

src/root.zig
```diff
@@ -71,6 +70,10 @@ export fn getenv(name_z: types.NullTerminatedString) ?types.NullTerminatedString
 }
 
 fn getEnvValue(name: [:0]const u8, configuration: config.InjectorConfiguration) ?types.NullTerminatedString {
+    if (!setCustomEnvVariables(configuration.all_auto_instrumentation_agent_env_vars)) {
+        print.printError("failed to set environment variables from configuration", .{});
+    }
```
basti1302: _issue:_ This code is run every time the process looks up an environment variable. I think it would make sense to memorize the fact that the custom env vars have already been injected after doing it for the first time.

src/root.zig
```diff
@@ -145,3 +148,96 @@ fn getEnvValue(name: [:0]const u8, configuration: config.InjectorConfiguration)
...
+            new_var_index += 1;
+            print.printDebug("added environment variable {s}={s}", .{ name, value });
+        }
+    }
+
+    // Add null terminator in the end
+    new_environment[new_var_index] = null;
+
+    const new_environ_ptr: [*]u8 = @ptrCast(new_environment.ptr);
+    __environ = new_environ_ptr;
```
basti1302, line 233: _note:_ I know you brought it up in the issue beforehand, but ultimately, overwriting `__environ` is not a feasible strategy, in the long run. The reason for that is that we need to get rid of declaring a dependency on `__environ`, that is, the actual problem is the declaration `extern var __environ: [*]u8;` in root.zig. So to be very clear, the problem is not with your changes in this PR but it is a more general problem with the current injector code base.

See https://github.com/dash0hq/dash0-operator/blob/0.91.0/images/instrumentation/injector/README.md#dynamically-linked-libc-free-binaries if you want to know why.

I think injecting the custom env vars like this is *fine for this PR*, based on the general approach that the injector uses at the moment. As I mentioned in the discussion in the issue, the whole approach will be revised, and then we will have to rework this part as well. The upside is that we will be able to use `setenv` then, so this will actually become a lot easier.

No need for you to change anything here, just a general note and additional context. :-)

test/apps/dotnet/App.cs
```diff
@@ -23,6 +23,9 @@ static int Main(string[] args)
             case "verify-startup-hook-has-been-injected":
                 EchoEnvVar("otel_injector_dotnet_no_op_startup_hook_has_been_loaded");
                 break;
+            case "custom-env-var":
+                EchoEnvVar("CUSTOM_ENV_VAR");
+                break;
```
basti1302, line 28: _praise:_ Thanks for covering the new feature also in the integration tests. 👍 👍

basti1302 (changes requested):
🎉 🎉 Excellent work! This looks very complete, has great test coverage, no major issues.

I have added a couple of comments inline, please take a look.

I marked one issue as blocking, but it is a really minor one, and I only marked it as blocking because it will help reduce merge conflicts with other PRs that are currently open..

src/config.zig
sultanowskii: Renamed

src/config.zig
sultanowskii: No, not really. I just wanted this function to be local - AFAIK there is no other way to do that in Zig.

Anyways, extracted the function 👍

packaging/fpm/etc/opentelemetry/otelinject.conf
sultanowskii: Thought about this one too, sound reasonable, changed to the `default_auto_instrumentation_env.conf`

src/config.zig
sultanowskii: Same as above.

Fixed

src/root.zig
sultanowskii: Thanks, fixed!

src/root.zig
sultanowskii: Fair, fixed

src/config.zig
sultanowskii: Fixed

src/config.zig
sultanowskii: Thanks for the explanation, renamed them back

sultanowskii:
@basti1302 Fixed issues, replied to your comments

Also, CI jobs are constantly failing during downloads - could you please restart (and monitor) them?

sultanowskii:
Aaand it failed again :(

basti1302:
Github recovered. Thanks again for the contribution!

src/config.zig
```diff
@@ -45,52 +53,98 @@ pub fn readConfiguration() InjectorConfiguration {
...
 }
 
+fn printErrorReturnEmptyConfig(err: std.fmt.AllocPrintError) InjectorConfiguration {
+    print.printError("Cannot allocate memory for the default injector configuration: {}", .{err});
+    return InjectorConfiguration{
+        .dotnet_auto_instrumentation_agent_path_prefix = "",
+        .jvm_auto_instrumentation_agent_path = "",
+        .nodejs_auto_instrumentation_agent_path = "",
+        .all_auto_instrumentation_agents_env_path = "",
+        .all_auto_instrumentation_agents_env_vars = std.StringHashMap([]u8).init(alloc.page_allocator),
```
jack-berg, line 68: Where does deinit get called on this?

src/config.zig
```diff
@@ -45,52 +53,98 @@ pub fn readConfiguration() InjectorConfiguration {
...
-    const nodejs_default =
-        std.fmt.allocPrint(alloc.page_allocator, "{s}", .{default_nodejs_auto_instrumentation_agent_path}) catch |err| {
-            print.printError("Cannot allocate memory for the default injector configuration: {}", .{err});
-            return InjectorConfiguration{
-                .dotnet_auto_instrumentation_agent_path_prefix = "",
-                .jvm_auto_instrumentation_agent_path = "",
-                .nodejs_auto_instrumentation_agent_path = "",
-            };
-        };
+    const all_agent_env_vars = std.StringHashMap([]u8).init(alloc.page_allocator);
```
jack-berg, line 73: Where does deinit get called on this?

Issue #84, opened 2025-09-15
Support opt-in and opt-out options


jackshirazi:
### Is your feature request related to a problem? Please describe.

With the operator, you have specific types of services deployed to a k8s namespace, so it's safe to auto-instrument them all. But we learned from our customers that they deploy all sorts of stuff to a system, including lots of utilities that definitely cannot cope with the agent being added in. We've had customers that for ease of deployment simply defined a global JAVA_TOOL_OPTIONS that always included the agent, and we had to add specific configurable opt-out and opt-in capabilities in the agent startup so that the agent never initialized for various cases.

These are exactly the type of customers that want a one-install auto-instrumentation capability like the injector - but at the same time it won't work for them because of this limitation that it would auto-instrument things that can't cope with being instrumented!

### Describe the solution you'd like

I think a similar approach to that suggested in [Support multiple agent versions](https://github.com/open-telemetry/opentelemetry-injector/issues/83) will work nicely! It would work for both opt-in and opt-out whichever the user prefers, eg another configuration file that uses a regex key to point to the appropriate conf file. So eg `otelmetainject.conf` with contents for opt-out

```
/no-instrumentation/=
default=/path/otelinject.conf
```

or for opt-in

```
/do-instrumentation/=/path/otelinject.conf
default=
```

where if the process command-line matches the regex key, it uses the corresponding config file, and no file means don't add any agent

### Describe alternatives you've considered

I think you would need to define `OTEL_INJECT_CONFIG_PATH` on a per-process basis (works for both opt-in and opt-out, but is cumbersome)

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

basti1302:
@jackshirazi I think we can close this, wdyt? #151 added such a mechanism.

PR #151, opened 2025-12-04
Add an options to include or exclude certain programs based on their executable path or command line arguments

grcevski:
I'm proposing a solution for https://github.com/open-telemetry/opentelemetry-injector/issues/84. Essentially, a way to include or exclude certain programs based on pattern matching on the program executable path and the program command line arguments.

I've added 4 new options:
```
include_paths
exclude_paths
include_with_arguments
exclude_with_arguments
```
If none of the options are set, the behaviour is identical to the existing behaviour, e.g. instrument everything if the language checks pass.

The options are additive if multiple lines are set, which means that the users can list each option on one line, by comma separating the values, or list each inclusion or exclusion criteria on a separate line.

I've also made exclusion criteria override the inclusion criteria, so that people can combine the different settings. I've provided some examples in the README documentation, of what kind of selections can be performed. For example, Java applications that run with small amount of heap (less than 10M maximum memory) can be avoided, or programs in certain directories, or programs that maybe load other agents.

For the pattern matching I opted for globs (widely used in OTel) and not bringing any external dependencies.

Relates to https://github.com/open-telemetry/opentelemetry-injector/issues/84

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -84,10 +84,10 @@ fn printErrorReturnEmptyConfig(err: std.fmt.AllocPrintError) InjectorConfigurati
         .nodejs_auto_instrumentation_agent_path = "",
         .all_auto_instrumentation_agents_env_path = "",
         .all_auto_instrumentation_agents_env_vars = std.StringHashMap([]u8).init(alloc.page_allocator),
-                .include_paths = &.{},
-                .exclude_paths = &.{},
-                .include_args = &.{},
-                .exclude_args = &.{},
+        .include_paths = &.{},
+        .exclude_paths = &.{},
+        .include_args = &.{},
+        .exclude_args = &.{},
     };
 }
 
@@ -113,10 +113,21 @@ fn createDefaultConfiguration() InjectorConfiguration {
         .nodejs_auto_instrumentation_agent_path = nodejs_default,
         .all_auto_instrumentation_agents_env_path = all_env_default,
         .all_auto_instrumentation_agents_env_vars = all_agent_env_vars,
-                .include_paths = &.{},
-                .exclude_paths = &.{},
-                .include_args = &.{},
-                .exclude_args = &.{},
+        .include_paths = &.{},
+        .exclude_paths = &.{},
+        .include_args = &.{},
+        .exclude_args = &.{},
+    };
+}
+
+fn applyCommaSeparatedPatternsOption(setting: *[][]const u8, value: []u8, pattern_name: []const u8, cfg_file_path: []const u8) void {
+    const new_patterns = patterns_util.splitByComma(alloc.page_allocator, value) catch |err| {
+        print.printError("error parsing {s} value from configuration file {s}: {}", .{ pattern_name, cfg_file_path, err });
+        return;
+    };
+    setting.* = std.mem.concat(alloc.page_allocator, []const u8, &.{ setting.*, new_patterns }) catch |err| {
+        print.printError("error concatenating {s} from configuration file {s}: {}", .{ pattern_name, cfg_file_path, err });
+        return;
     };
 }
 
@@ -129,6 +140,14 @@ fn applyKeyValueToGeneralOptions(key: []const u8, value: []u8, _cfg_file_path: [
         _configuration.jvm_auto_instrumentation_agent_path = value;
     } else if (std.mem.eql(u8, key, nodejs_path_key)) {
         _configuration.nodejs_auto_instrumentation_agent_path = value;
+    } else if (std.mem.eql(u8, key, include_paths_key)) {
+        applyCommaSeparatedPatternsOption(&_configuration.include_paths, value, "include_paths", _cfg_file_path);
+    } else if (std.mem.eql(u8, key, exclude_paths_key)) {
+        applyCommaSeparatedPatternsOption(&_configuration.exclude_paths, value, "exclude_paths", _cfg_file_path);
+    } else if (std.mem.eql(u8, key, include_args_key)) {
+        applyCommaSeparatedPatternsOption(&_configuration.include_args, value, "include_arguments", _cfg_file_path);
+    } else if (std.mem.eql(u8, key, exclude_args_key)) {
+        applyCommaSeparatedPatternsOption(&_configuration.exclude_args, value, "exclude_arguments", _cfg_file_path);
     } else {
         print.printError("ignoring unknown configuration key in {s}: {s}={s}", .{ _cfg_file_path, key, value });
         alloc.page_allocator.free(value);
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -30,7 +30,6 @@ See the page [Zig Learn](https://ziglang.org/learn/) for more information about
 * `make zig-unit-tests` to run the unit tests
 * `watch-zig-unit-tests` to continually run the Zig unit tests on every change (requires [`fd`](https://github.com/sharkdp/fd) and [`entr`](https://github.com/eradman/entr) to be installed)
 * `make tests` to run the unit tests and the injector integration tests
-* `make tests` to run unit tests and the injector integration tests
```
atoulme: can be a separate PR

grcevski: Reverted, I'll submit a separate PR for this docs change. Thanks!

grcevski: Separate PR: https://github.com/open-telemetry/opentelemetry-injector/pull/154

src/args_parser.zig
```diff
@@ -0,0 +1,47 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+const std = @import("std");
+
+/// Parses /proc/<pid>/cmdline and extracts the executable and arguments.
+/// Returns the executable path and a slice of arguments (excluding the executable).
```
skl: This comment says the executable is excluded, but the comment on line 45 seems to indicate that it is included. Which is correct?

src/patterns_util.zig
```diff
@@ -0,0 +1,76 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+const std = @import("std");
+const alloc = @import("allocator.zig");
+const print = @import("print.zig");
+const testing = std.testing;
+
+/// Splits a comma-separated string into a slice of trimmed strings.
+pub fn splitByComma(input: []const u8) ![][]const u8 {
+    var list: std.ArrayList([]const u8) = .init(alloc.page_allocator);
```
jaronoff97: Why use the global allocator here, but pass the allocator elsewhere?

README.md
```diff
@@ -51,24 +51,95 @@ This method requires `root` privileges.
    jvm_auto_instrumentation_agent_path=/usr/lib/opentelemetry/javaagent.jar
    nodejs_auto_instrumentation_agent_path=/usr/lib/opentelemetry/otel-js/node_modules/@opentelemetry-js/otel/instrument
    ```
-   There is usually no need to modify this file, unless you want to provide your own instrumentation files.
+   You may want to modify this file for a couple of reasons:
+   - You want to provide your own instrumentation files.
```
jaronoff97, line 55: I appreciate the level of detail here, however, it's pretty hard to understand what a user needs to do now. Can we instead provide a separate section with usage information for the new fields?

src/patterns_util.zig
grcevski: Good point, I will make this consistent. I realise I left some of this like this, because I was temporarily using the debug allocator to make sure I didn't leak anything while developing.

README.md
grcevski: Sure thing, I'll move this text into a separate section on excluding/including details.

src/args_parser.zig
grcevski: I had an earlier version which returned the executable separately but it was harder to use and deallocate the memory. I forgot to update the comment after making return everything in the same slice. Thanks for catching this. Fixed now.

src/patterns_util.zig
grcevski: I added the allocator in splitByComma to make it consistent with the rest of the code and pluggable for other allocators for leaks debugging.

README.md
grcevski: Moved the detailed text in a separate section.

src/args_parser.zig
```diff
@@ -0,0 +1,47 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+const std = @import("std");
+
+/// Parses /proc/<pid>/cmdline and extracts the executable and arguments.
+/// Returns the slice of the arguments including the executable as the first argument.
+/// The cmdline file contains null-separated arguments.
+/// Caller owns the returned memory and must free it.
+pub fn cmdLineForPID(allocator: std.mem.Allocator) ![]const []const u8 {
```
basti1302, line 10: _thought:_ Having this facility in the injector will be very handy for reasons completely unrelated to the new configuration settings. I would very much like to include the executable name (and the pid) in the log messages the injector emits, for the following reasons: when instrumented processes start child processes, the child process will also be injected. You might see multiple log lines being emitted to stderr without realizing that they actually come from different processes (say `npm start` -> `node index.js` and similar scenarios).

Note that this is not a call to action for changing the logging in this PR, this is certainly out of scope for this PR.

src/patterns_matcher.zig
```diff
@@ -0,0 +1,213 @@
...
+    try testing.expect(!matchGlob("*.txt", "file.log"));
+}
+
+test "matchGlob: empty pattern and text" {
+    try testing.expect(matchGlob("", ""));
+    try testing.expect(!matchGlob("", "text"));
+    try testing.expect(!matchGlob("pattern", ""));
+}
+
+test "matchGlob: star matches empty" {
```
basti1302: nit: I think the test name and test content do not really match (the test does not only test * vs empty string).

src/patterns_matcher.zig
```diff
@@ -0,0 +1,213 @@
...
+    const patterns: []const []const u8 = &.{ "/usr/*/ba?h", "*.conf", "/opt/*/*" };
+    const args: []const []const u8 = &.{ "/usr/bin/myprogram", "/usr/local/bash", "script.sh" };
+    try testing.expect(matchesManyAnyPattern(args, patterns));
+}
+
+test "matchesManyAnyPattern: program name not matched" {
+    const patterns: []const []const u8 = &.{"/usr/bin/myprogram"};
+    const args: []const []const u8 = &.{ "/usr/bin/myprogram", "arg1" };
+    try testing.expect(!matchesManyAnyPattern(args, patterns));
+}
```
basti1302, line 213: _praise:_ I love the comprehensive test coverage for this. 🚀

src/args_parser.zig
```diff
@@ -0,0 +1,47 @@
...
+    const cmdline_path = "/proc/self/cmdline";
+
+    const file = std.fs.openFileAbsolute(cmdline_path, .{}) catch |err| {
```
basti1302: _thought:_ A pattern that we use in other parts of the code is to split a function that uses a well-known file path (like `/proc/self/cmdline` here) into two parts:
- the actual implementation, which accepts a file path as an argument (say `getCmdLineForPID`) and
- a very short `pub` function, say `pub fn cmdLineForPID` that then only delegates to `getCmdLineForPID` and hands over the file path argument (i.e. `pub fn cmdLineForPID` basically only calls `getCmdLineForPID("/proc/self/cmdline")`)

Compare for example `pub fn checkOTelJavaAgentJarAndGetModifiedJavaToolOptionsValue` vs  `doCheckOTelJavaAgentJarAndGetModifiedJavaToolOptionsValue`.

That way, the actual implementation `getCmdLineForPID` can be unit tested, e.g. with an invalid/non-existing path or even with actual files from `unit-test-assets`.

This is not blocking though, the test coverage for this is already really good I think.

README.md
```diff
@@ -51,24 +51,33 @@ This method requires `root` privileges.
...
+      dotnet_auto_instrumentation_agent_path_prefix=
+      jvm_auto_instrumentation_agent_path=/usr/lib/opentelemetry/javaagent.jar
+      nodejs_auto_instrumentation_agent_path=/usr/lib/opentelemetry/otel-js/node_modules/@opentelemetry-js/otel/instrument
+      ```
+   - You want to selectively enable (or disable) auto-instrumentation for a subset of programs (services) on your system. 
+     For example, you may want to only enable instrumentation of services that match a specific executable path pattern, or
+     to programs that do not contain certain arguments on the command line. 
+     See [details on configuring the program inclusion and exclusion criteria](#details-on-configuring-the-program-inclusion-and-exclusion-criteria) for more information.
 
    The paths set in `/etc/opentelemetry/otelinject.conf` can be overridden with environment variables.
```
basti1302: _super-nit:_ We might want to change "The paths set in" to "The values set in"; it is not only paths anymore.

README.md
```diff
@@ -113,6 +122,73 @@ Check the following for details about the auto-instrumtation agents and further
 - [Node.js](https://opentelemetry.io/docs/zero-code/js/configuration/)
 - [.NET](https://opentelemetry.io/docs/zero-code/dotnet/configuration/)
 
+### Details on configuring the program inclusion and exclusion criteria
+
+If you want to selectively enable (or disable) auto-instrumentation for a subset of programs (services) on your system, 
+the configuration file provides a couple of settings which can be used alone or in combination to produce 
+the desired outcome:
+  - `include_paths` - A comma-separated list of glob patterns to match executable paths. 
+    If you do not specify anything here, the injector defaults to instrumenting **all eligible executables**.
```
basti1302: _question:_ should we remove "eligible" here? It really defaults to instrumenting *all* executables, technically speaking, doesn't it? (Although it will not do much unless the executable asks for NODE_OPTIONS or similar.)

If we leave it in, we might want to/need to explain what "eligible" means in this context?

README.md
```diff
@@ -113,6 +122,73 @@ Check the following for details about the auto-instrumtation agents and further
...
+  - `exclude_paths` - A comma-separated list of glob patterns to exclude executable paths.
+  - `include_with_arguments` - A comma-separated list of glob patterns to match process arguments.
+    If you do not specify anything here, the injector defaults to instrumenting **all eligible executables**.
```
basti1302: same question about "eligible" here

README.md
```diff
@@ -113,6 +122,73 @@ Check the following for details about the auto-instrumtation agents and further
...
+  - any programs that run from the `/app` and `/utilities` directories
+  - any programs that have an `.exe` extension
+  - any programs that have a command line argument containing `MyProject` 
+  in the name and ending with the extension `.jar`
+  - any programs that have a program argument ending in `app.js`
+  - however, for all included programs, the injector **will avoid**: 
+    - all programs that have a command line argument starting with
+    `-javaagent:`
+    - all programs that have a command line argument that contains `@opentelemetry-js`
+    - all `java` programs that run with a single digit megabytes of maximum memory
```
basti1302, line 182: _praise:_ love the real world use case examples! Super helpful.

README.md
grcevski: Good point.

README.md
grcevski: done

README.md
grcevski: fixed.

README.md
grcevski: fixed

src/args_parser.zig
grcevski: Great idea! I've split the function now and I was easily able to write new tests to prove the parsing of the content works well.

src/patterns_matcher.zig
grcevski: fixed.

src/args_parser.zig
grcevski: Happy to follow up with a PR to add some logging for this.

grcevski:
> @grcevski Thanks for resolving all the review comments. I think this is very ready to be merged now. With #148 now being merged there are two smaller conflicts. Would you mind rebasing this on `main`? Thanks so much in advance.

Done! Thanks!

src/args_parser.zig
basti1302: @grcevski do you still plan to create that follow up PR around adding the pid and executable name to the log output? 😊😇

grcevski: Yeah sorry, memory loss after being off for more than 2 weeks :). I will get to this on Monday or Tuesday.

basti1302: Awesome! No worries, no need to rush this. 👍

grcevski: I made a PR here: https://github.com/open-telemetry/opentelemetry-injector/pull/200

PR #152, opened 2025-12-04
[chore] upgrade Zig to 0.15.2

inge4pres:
### Reason for this PR

Zig 0.15.2 is the most recent stable version.
Update the codebase to use it.

### Details

* update managed ArrayList to unmanaged
* `allocPrintZ` changed to `allocPrintSentinel`
* use std.Io.Reader in `readConfigurationFile`

Bonus: add a `-Dtest-filter` option to build.zig to allow running unit tests with filters.

### Additional context

Running `make tests` is recommended but it's failing with a weird error on permissions 🤔 
```
Unable to find image 'libotelinject-builder:amd64' locally
docker: Error response from daemon: pull access denied for libotelinject-builder, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

I tried running  `make dist`, it seems the right one to build the image but on Linux apparently the local images aren't resolved 🤷🏼 

```
docker run -d --platform linux/amd64 --name libotelinject-builder libotelinject-builder:amd64 sleep inf
Unable to find image 'libotelinject-builder:amd64' locally
docker: Error response from daemon: pull access denied for libotelinject-builder, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -76,8 +76,8 @@ pub fn readConfiguration() InjectorConfiguration {
     return configuration;
 }
 
-fn printErrorReturnEmptyConfig(err: std.fmt.AllocPrintError) InjectorConfiguration {
-    print.printError("Cannot allocate memory for the default injector configuration: {}", .{err});
+fn printErrorReturnEmptyConfig(err: std.mem.Allocator.Error) InjectorConfiguration {
+    print.printError("Cannot allocate memory for the default injector configuration: {t}", .{err});
     return InjectorConfiguration{
         .dotnet_auto_instrumentation_agent_path_prefix = "",
         .jvm_auto_instrumentation_agent_path = "",
@@ -156,7 +156,10 @@ fn applyKeyValueToGeneralOptions(key: []const u8, value: []u8, _cfg_file_path: [
 
 fn readConfigurationFile(cfg_file_path: []const u8, configuration: *InjectorConfiguration) void {
     const config_file = std.fs.cwd().openFile(cfg_file_path, .{}) catch |err| {
-        print.printDebug("The configuration file {s} does not exist or cannot be opened. Configuration will use default values and environment variables only. Error: {}", .{ cfg_file_path, err });
+        print.printDebug(
+            "The configuration file {s} does not exist or cannot be opened. Configuration will use default values and environment variables only. Error: {t}",
+            .{ cfg_file_path, err },
+        );
         return;
     };
     defer config_file.close();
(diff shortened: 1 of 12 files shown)


src/config.zig
```diff
@@ -87,58 +87,46 @@ fn createDefaultConfiguration() InjectorConfiguration {
...
-                continue;
+                print.printError(
+                    "A line in configuration file {s} exceeds the maximum buffer length of {d}",
+                    .{ cfg_file_path, max_buffer_len },
+                );
             },
-            else => |e| {
-                print.printError("read error while reading configuration file {s}: {}", .{ cfg_file_path, e });
-                break;
+            // if the file does not end with a newline, we still need to parse the last line
```
atoulme: that's a change in behavior

src/test_util.zig
```diff
@@ -35,11 +35,7 @@ pub fn setStdCEnviron(env_vars: []const []const u8) anyerror![*:null]?[*:0]u8 {
     // allocator.allocSentinel(?[*:0]u8, n, null).
     const new_environ = try test_allocator.allocSentinel(?[*:0]u8, env_vars.len, null);
     for (env_vars, 0..) |env_var, i| {
-        new_environ[i] = try std.fmt.allocPrintZ(
-            test_allocator,
-            "{s}",
-            .{env_var},
-        );
+        new_environ[i] = try std.fmt.allocPrintSentinel(test_allocator, "{s}", .{env_var}, 0);
```
atoulme: Is there a reason to change the indent?

build.zig
```diff
@@ -71,6 +71,7 @@ pub fn build(b: *std.Build) !void {
     b.default_step = copy_injector_to_bin;
 
     // TESTING
+    const test_filters = b.option([][]const u8, "test-filter", "Match tests to execute");
```
atoulme, line 74: Is that a change that can be offered in a separate PR?

zig-version
```diff
@@ -1,3 +1,3 @@
 # See https://dl-cdn.alpinelinux.org/alpine/edge/community, check for zig-*.apk for available versions.
-# Keep this in sync with .github/workflows/ci.yaml -> mlugg/setup-zig.with.version.
-ZIG_VERSION=0.14.1-r0
+# Keep this in sync with .github/workflows/build.yaml -> mlugg/setup-zig.with.version.
```
atoulme, line 2: Is this a typo in the comment?

src/config.zig
inge4pres: Are you referring to the error message or the parsing of the last line?
The error previously grouped by the `else` prong is now `ReadFailed` using the new API.
The parsing of last line has a different internal behavior, because `EndOfStream` is returned when the file ends without a newline, and if the the last line is longer than `max_line_length` we discard it.

The unit tests are passing, we are not covering the case where there's a line that exceeds the increased read buffer returning `StreamTooLong`, let me know if I should add it.

src/test_util.zig
inge4pres: I guess i forgot a comma after the 0, the formatting is done automatically by `zig fmt`.
I will change it back.

build.zig
inge4pres: Yes, I'll remove the commit that adds this and put it in a separate PR.

zig-version
inge4pres: I believe so. There is no `ci.yaml` file in the .github/workflows folder.
The version needs to be in sync with wjhat'sin build.yaml.

I would leave this type fix in this PR along the version upgrade, let me know.

basti1302: Yes, this was a typo, thanks for catching it. 👍  (It was ci.yaml in the Dash0 repo I sourced large parts of the code from.)

basti1302: And for the record, I really don't think it is worth splitting things like this off into separate PRs.

src/config.zig
basti1302: If the unit tests are passing then I think we are good with respect to this update not changing the behavior. 

> The parsing of last line has a different internal behavior, because EndOfStream is returned when the file ends without a newline, and if the the last line is longer than max_line_length we discard it.

Let me rephrase this to see if I got this right: This is only about the edge case that the last line is overly long _and_ does not end in a newline right?

From my point of view, we do not need to cover this additional edge case in this PR, since it was also not covered before.

inge4pres: > This is only about the edge case that the last line is overly long and does not end in a newline right?

Not exactly. The previous Reader API didn't need an user-provided buffer, it used to use an internal one sized with `max_line_length`; the new std.Io.Reader API requires to provide a buffer.
I decided to set it twice `max_line_length` (8192 chars * 2) to support the case of reading up to 8192 chars.
With this PR, when iterating through the file, regardless of the position, if we try to read more than 8192 * 2 we get `error.StreamTooLong`, because we exceed the size of the user-provided buffer.
In this case, we return a _new_ error message that is added specifically to handle it at L118

```zig
                print.printError(
                    "A line in configuration file {s} exceeds the maximum buffer length of {d}",
                    .{ cfg_file_path, max_buffer_len },
                );
```
this is not covered by an additional test in this PR.

The existing behavior of discarding lines longer than 8192 is preserved with the if branch at [L103](https://github.com/inge4pres/opentelemetry-injector/blob/1380555d2924d46362caf4529bb9cce98569058b/src/config.zig#L103-L109).

basti1302: Got it, thanks for the detailed explanation. I think covering this new case is not strictly necessary, but would be beneficial. (I.e. not adding it is not blocking this PR).

inge4pres:
> @inge4pres Now that the other PRs have been merged, we should get this one merged. Could you rebase on `main`? There might be some new code that needs to be lifted to 0.15.2, not sure. Thanks in advance!

Rebased and ran build and tests locally with success 🎉

src/config.zig
```diff
@@ -153,10 +153,36 @@ fn applyKeyValueToGeneralOptions(key: []const u8, value: []u8, _cfg_file_path: [
...
+//     }
+
+//     const env_file = std.fs.cwd().openFile(env_file_path, .{}) catch |err| {
+//         print.printDebug("The configuration file {s} does not exist or cannot be opened. Error: {}", .{ env_file_path, err });
+//         return;
+//     };
+//     defer env_file.close();
+
+//     return parseConfiguration(configuration, env_file, env_file_path, applyKeyValueToAllAgentsEnv);
+// }
```
basti1302: @inge4pres This looks like an artifact from merging/rebasing which was committed accidentally.

inge4pres: ah apologies! fixed

inge4pres:
> Thanks so much for your patience.

Don't mention it, glad we upgrade!

PR #158, opened 2025-12-13
build: use enum to define cpu options

inge4pres:
### Reason for this PR

Cosmetic change to `build.zig` to define the `cpu-arch` build option as an enum.

### Details

* Introduce a `SupportedCpuArch` enum with the existing CPU variants
* Attach methods to get cou arch and model

## Additional context

Benefit: the enum might be propagated in other parts of the code where switching on the CPU is needed (e.g. src/dotnet.zig L331)
Downside: the error message for unsupported architecture is now a little less clear.

```
☁  opentelemetry-injector [build-cpu-enum] ⚡  zig build -Dcpu-arch=blah
error: Expected -Dcpu-arch to be of type build.SupportedCpuArch.
error:   access the help menu with 'zig build -h'
error: the following build command failed with exit code 1:
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -3,32 +3,17 @@
 
 const std = @import("std");
 
-pub const InjectorBuildError = error{UnsupportedArchitecturError};
-
 // Although this function looks imperative, note that its job is to
 // declaratively construct a build graph that will be executed by an external
 // runner.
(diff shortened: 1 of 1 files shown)


basti1302 (approved):
Nice!

PR #166, opened 2025-12-15
feat: remove link-time dependency on __environ, inject via setenv

basti1302:
(1) This removes the declaration
    extern var __environ: [*]u8;
from the injector, that is, the dependency of the __environ pointer symbol exported by libc.

The injector may be linked to binaries that, while dynamically linked, do not link any libc. This would result in the injector preventing the program from starting due to a linker error, effectively crashing the application under monitoring.

This change introduces a way to look up the dlsym symbol, and then in turn through dlsym the __environ and setenv symbols. This happens at runtime in an initialization step, by inspecting the process' memory and finding the dlsym symbol in memory. This allows us to read and manipulate the process environment without depending on any external symbols at link time.

The way we do this is in three steps:
1. Find out which libc the process has (if any) -- we already did read the ELF metadata (DT_NEEDED) as some OpenTelemetry SDKs need to be injected differently (using different binaries) depending on the libc flavor.
2. In the loaded libc, find the location of the `dlsym` function reading ELF metadata. To accomplish this, the change "forks" the ElfDynLib from `dynamic_library` and modifies it to remove stuff we don’t care about (e.g., symbol version matching), and to prevent it from memory mapping (as what we need was already mapped in memory by the process loading); think of it as “just enough Elf to find a symbol the process already has”; as a side note, it’s fascinating that, by casting memory locations, we can achieve the lookup without a single allocation.
3. Let the libc's `dlsym` handle the fifty quintillion corner cases (aliases, lazy loading, linker resolution order, `__environ` exported by the main binary, etc.) and find the `__environ` and `setenv` symbols.

(2) The injector now no longer overrides libc's getenv to inject environment variables but instead unconditionally injects all environment variables by calling setenv at startup. This change is necessary, because even when removing the dependency on the __environ symbol (see above), there would still be potential to break certain executables:
- For example, on some older distributions (Debian 11/bullseye for example), libc and libdl are two distinct libary files, with libdl containing dlopen, dlcose, dlsym and dlerror. When a binary only binds libc, but not libdl, we would not find the dlsym symbol (because it is actually really not loaded into memory), hence we wouldn't be able to find the __environ or setenv symbols. As a consequence, we would not be able to backfill the injector-internal representation of the environment (std.os.environ/std.posix.getenv in the Zig code) and _all_ calls to getenv would return null; i.e. we would effectively obliterate the environment of the process.
- Another issue is that, if a binary binds libcrypto (i.e. uses OpenSSL), it will try to look up the OPENSSL_armcap environment variable before the injector init code even had ran. With the getenv override, this would always return null, even if OPENSSL_armcap was set (independent of the libdl issue described above). Hence, overriding getenv would break some of OpenSSL's assumptions.
- Many runtimes do not use getenv to read environment variables. E.g. the JVM as well as the CLR use getenv in their native startup code, but when environment variables are read from with the runtime, the reading bypasses getenv. Hence, overriding getenv is not feasible to inject reliably into these runtimes. The same is true for Python. Injecting via setenv solves this problem.

(3) Additional changes in this commit:

- The contract of getEnvVarValue has changed: it now returns null when the injector does not need to make a change to an environment variable (i.e. because the original value is already correct). This is possible because we do no longer override libc's getenv, so it is not necessary to provide the environment variable's original value, we can just decide to not call setenv for this variable
- We no longer inject resource attributes into the JVM via the workaround of adding -Dotel.resource.attributes to JAVA_TOOL_OPTIONS. This was only necessary because the JVM does not read OTEL_RESOURCE_ATTRIBUTES via getenv, hence this was a necessary workaround with the override-getenv approach. With the switch to setenv, it is no longer necessary.

Commit authorship/credits:
* This is sourced from https://github.com/dash0hq/dash0-operator/pull/490
* The code for parsing the ELF metadata, parsing /proc/self/maps, looking up dlsym, setenv and __environ from the host process' memory regions (that is, elf.zig, auxv.zig, and libc.zig) have been authored by Michele Mancioppi in the upstream repository (dash0-operator). (Since this is the core change that enables everything else, this commit is tagged with Michele as the main author.)
* The change from the getenv-override to setenv, additional fixes for looking up symbols in memory, and a good chunk of cleanup and test code has been contributed by Bastian Krol.
* This commit in the opentelemetry-injector repository was authored by Bastian, but includes Michele's contributions from the downstream repository.

Co-authored-by: Michele Mancioppi <michele.mancioppi@dash0.com>
Co-authored-by: Bastian Krol <bastian.krol@dash0.com>


src/allocator.zig
```diff
@@ -3,6 +3,4 @@
 
 const std = @import("std");
 
-// We need to allocate memory only to manipulate and return the few environment variables we want to modify. Unmodified
-// values are returned as pointers to the original `__environ` memory.
```
basti1302, line 7: No longer accurate, we to not return unmodified env vars anymore (because we do not override getenv anymore), we simply do not `setenv` unmodified env vars.

src/dotnet.zig
```diff
@@ -6,30 +6,13 @@ const std = @import("std");
...
-// https://github.com/dotnet/runtime/blob/v9.0.5/src/native/corehost/hostpolicy/shared_store.cpp#L16
-// -> https://github.com/dotnet/runtime/blob/v9.0.5/src/native/corehost/hostmisc/pal.unix.cpp#L954.
-//
-// In contrast to that, the implementation of Environment.GetEnvironmentVariable reads __environ into a dictionary
-// and then uses that dictionary for all lookups, see here:
-// https://github.com/dotnet/runtime/blob/v9.0.5/src/libraries/System.Private.CoreLib/src/System/Environment.cs#L66 ->
-// - https://github.com/dotnet/runtime/blob/v9.0.5/src/libraries/System.Private.CoreLib/src/System/Environment.Variables.Unix.cs#L15-L32,
-// - https://github.com/dotnet/runtime/blob/v9.0.5/src/libraries/System.Private.CoreLib/src/System/Environment.Variables.Unix.cs#L85-L91, and
-// https://github.com/dotnet/runtime/blob/v9.0.5/src/libraries/System.Private.CoreLib/src/System/Environment.Variables.Unix.cs#L93-L166
-
```
basti1302, line 32: With the switch back to setenv, the problem described above becomes a non-issue.

src/dotnet.zig
```diff
@@ -40,50 +23,76 @@ pub const DotnetValues = struct {
...
+// root.zig#initEnviron for eaech .NET-related env var. This is simply because .NET requires multiple environment
+// variables to be set.
+var cached_dotnet_values = CachedDotnetValues{
+    .values = null,
+    .done = false,
+};
+var libc_flavor: ?types.LibCFlavor = null;
 
-const injection_happened_msg = "injecting the .NET OpenTelemetry instrumentation";
-var injection_happened_msg_has_been_printed = false;
```
basti1302, line 55: The `injecting the ...` messages have been removed in favor of logging this in a central place in root.zig#modifyEnvironmentVariable. (`setting "CORECLR_ENABLE_PROFILING"="1"` etc., at `debug` level)

src/dotnet.zig
```diff
@@ -96,240 +105,58 @@ fn doGetDotnetValues(dotnet_path_prefix: []u8) ?DotnetValues {
...
 
     unreachable;
 }
 
-test "doGetDotnetValues: should return null value if the profiler path cannot be accessed" {
-    const dotnet_values = doGetDotnetValues(try std.fmt.allocPrint(test_util.test_allocator, "/invalid/path", .{}));
-    try testing.expect(dotnet_values == null);
-}
-
-fn getLibCFlavor() LibCFlavor {
```
basti1302, line 116: The code to determine the libc flavor was only relevant for .NET previously, which is why it was here in dotnet.zig. This is now all in libc.zig, since it is relevant for the injector as a whole now.

src/jvm.zig
```diff
@@ -30,414 +29,95 @@ fn doCheckOTelJavaAgentJarAndGetModifiedJavaToolOptionsValue(
     jvm_auto_instrumentation_agent_path: []u8,
 ) ?types.NullTerminatedString {
     if (jvm_auto_instrumentation_agent_path.len == 0) {
-        print.printMessage("Skipping the injection of the OpenTelemetry Java agent in \"JAVA_TOOL_OPTIONS\" because it has been explicitly disabled.", .{});
-        if (original_value_optional) |original_value| {
-            return original_value;
-        }
```
basti1302, line 36: The contract between root.zig and the "xxxGetModifiedValue" functions in node_js.zig/jvm.zig has changed: When xxxGetModifiedValue returns null, root.zig will simply not call setenv for this key value pair. This allows returning null from the xxxGetModifiedValue to signal to root.zig in all situations where we do not want to change an env var (either because it already has the correct value or because we have encountered an error).

This is why the branches with `return original_value;` are no longer necessary.

src/jvm.zig
```diff
@@ -30,414 +29,95 @@ fn doCheckOTelJavaAgentJarAndGetModifiedJavaToolOptionsValue(
...
     const javaagent_flag_value = std.fmt.allocPrintSentinel(alloc.page_allocator, "-javaagent:{s}", .{jvm_auto_instrumentation_agent_path}, 0) catch |err| {
         print.printError("Cannot allocate memory to manipulate the value of \"{s}\": {}", .{ java_tool_options_env_var_name, err });
-        if (original_value_optional) |original_value| {
-            return original_value;
-        }
         return null;
     };
 
-    const original_otel_resource_attributes_env_var_value_optional =
-        std.posix.getenv(res_attrs.otel_resource_attributes_env_var_name);
```
basti1302, line 59: We no longer need the clutch of setting OTEL_RESOURCE_ATTRIBUTES by adding -Dotel.resource.attribtues to JAVA_TOOL_OPTIONS for JVMs, so the whole business of managing resource attributes in jvm.zig can be removed.

This makes jvm.zig a lot simpler, and it is the reason why a lot of cases have been removed below.

src/print.zig
```diff
@@ -3,40 +3,225 @@
 
 const std = @import("std");
 
-const otel_injector_debug_env_var_name = "OTEL_INJECTOR_DEBUG";
```
basti1302, line 6: OTEL_INJECTOR_DEBUG is no longer supported. With the switch to setenv, our logging behavior is now a much more sensitive issue, since we might emit log lines to stderr on _every invocation of any executable_ on the system. This required introducing a more fine grained distinction of log levels, and setting the default log level to `error`.

src/print_test.zig
```diff
@@ -1,40 +0,0 @@
...
-
-const print = @import("print.zig");
-const test_util = @import("test_util.zig");
-
-const testing = std.testing;
-
-// Note on having tests embedded in the actual source files versus having them in a separate *_test.zig file: Proper
-// pure unit tests are usually directly in the source file of the production function they are testing. More invasive
-// tests that need to change the environment variables (for example) should go in a separate file, so we never run the
-// risk of even compiling the test mechanism to modify the environment.
```
basti1302, line 14: Since the log level is no longer based on the actual environment but on /proc/self/environ, we no longer need a separate test file. The logic to read /proc/self/environ and determine the log level based on that is unit-tested within print.zig.

src/resource_attributes.zig
```diff
@@ -56,131 +58,136 @@ const mappings: [8]EnvToResourceAttributeMapping =
             .environement_variable_name = "OTEL_INJECTOR_SERVICE_NAMESPACE",
             .resource_attributes_key = "service.namespace",
         },
-
-        // General purpose mapping from a comma-separated list (OTEL_INJECTOR_RESOURCE_ATTRIBUTES) to individual
-        // resource attributes.
-        EnvToResourceAttributeMapping{
-            .environement_variable_name = "OTEL_INJECTOR_RESOURCE_ATTRIBUTES",
-            .resource_attributes_key = null,
-        },
```
basti1302, line 65: This is now handled without having `OTEL_INJECTOR_RESOURCE_ATTRIBUTES` here in `mappings`, we just read `OTEL_INJECTOR_RESOURCE_ATTRIBUTES` explicitly in `getModifiedOtelResourceAttributesValue`.

src/root.zig
```diff
@@ -15,83 +16,127 @@ const types = @import("types.zig");
...
 const args_parser = @import("args_parser.zig");
 
-const assert = std.debug.assert;
-const testing = std.testing;
-const expect = testing.expect;
+const empty_z_string = "\x00";
 
-// We need to use a rather "innocent" type here, the actual type involves
-// optionals that cannot be used in global declarations.
-extern var __environ: [*]u8;
```
basti1302, line 24: All of this, just to get rid of this single line 😬 😀

src/root.zig
```diff
@@ -15,83 +16,127 @@ const types = @import("types.zig");
...
+    } else {
+        return error.CannotFindEnvironSymbol;
+    }
+}
+
+fn evaluateAllowDeny(configuration: config.InjectorConfiguration) bool {
+    const exe_path = getExecutablePath() catch {
+        // Skip allow-deny evaluation if getting the executable path has failed. The error has already been logged in
+        // getExecutablePath.
+        return true;
```
basti1302, line 134: Debatable: Previously, when we could not read the executable path or the args, we would still compare the allow-deny settings from the config with empty strings, which is somewhat pointless.

Now we allow injection into the processes if we cannot read the executable name or args. The only other reasonable behavior IMO would be to categorically stand down and not inject if we cannot read executable path or args.

Not entirely sure what is better. I think I would like to leave it like this for now until we see real-world scenarios where looking up the executable name does not work, then go from there.

src/root.zig
```diff
@@ -126,214 +171,134 @@ fn getEnvValue(name: [:0]const u8, configuration: config.InjectorConfiguration)
...
-    // returning a pointer to it.
-    if (original_value) |val| {
-        return val.ptr;
-    }
-
-    // The requested environment variable is not one that we want to modify, and it does not exist. Return null.
     return null;
 }
 
-fn setCustomEnvVariables(custom_env_vars: std.StringHashMap([]u8)) bool {
```
basti1302, line 208: This is now a lot easier due to the switch back to setenv.

src/test_util.zig
```diff
@@ -48,7 +50,7 @@ pub fn setStdCEnviron(env_vars: []const []const u8) anyerror![*:null]?[*:0]u8 {
 }
 
 /// Resets std.c.environ to the given value. This function is meant to be used after doing clearStdCEnviron or
-/// setStdCEnviron earler, to restore the original environment after the test is done.
+/// setStdCEnviron earlier, to restore the original environment after the test is done.
```
basti1302, line 53: No, I will not extract this typo-fix into a separate PR, thank you very much 😉

test/scripts/default.tests
```diff
@@ -27,7 +27,7 @@ run_test_case \
   "sets k8s.pod.uid and k8s.container.name via OTEL_RESOURCE_ATTRIBUTES with pre-existing value" \
   "app" \
   "node index.js otel-resource-attributes" \
-  "OTEL_RESOURCE_ATTRIBUTES: k8s.namespace.name=my-namespace,k8s.pod.name=my-pod,k8s.pod.uid=275ecb36-5aa8-4c2a-9c47-d8bb681b9aff,k8s.container.name=test-app,foo=bar" \
+  "OTEL_RESOURCE_ATTRIBUTES: foo=bar,k8s.namespace.name=my-namespace,k8s.pod.name=my-pod,k8s.pod.uid=275ecb36-5aa8-4c2a-9c47-d8bb681b9aff,k8s.container.name=test-app" \
```
basti1302, line 30: We now append otel resource attributes to the existing list, instead of prepending them. This is because the logic is now smart enough to not override existing key-value pairs. Previously we did that implicitly by prepending and relying on OTel SDKs to behave spec-compliant and give priority to the key-value pair that appears later in the list. The new way is much safer and easier to reason about.

README.md
```diff
@@ -145,21 +134,69 @@ Check the following for details about the auto-instrumtation agents and further
...
+  attributes. These environment variables need to be externally (for example by a Kubernetes operator with a mutating
+  webhook on the pod spec template of the workload). If `OTEL_RESOURCE_ATTRIBUTES` is already set, the additional
+  key-value pairs are appended to the existing value of `OTEL_RESOURCE_ATTRIBUTES`. Existing key-value pairs are not
+  overwritten, that is if e.g. `OTEL_RESOURCE_ATTRIBUTES` already has a key-value pair for `k8s.pod.name`, the existing
+  key-value pair takes priority.
+  The following environment variables and resource attributes are supported:
+    * `OTEL_INJECTOR_NAMESPACE_NAME` will be translated to `k8s.namespace.name`
+    * `OTEL_INJECTOR_POD_NAME` will be translated to `k8s.pod.name`
+    * `OTEL_INJECTOR_POD_UID` will be translated to `k8s.pod.uid`
+    * `OTEL_INJECTOR_CONTAINER_NAME` will be translated to `k8s.container.name`
```
jaronoff97: Maybe this is personal preference, but would it make sense for this to be `OTEL_INJECTOR_K8S_*`?

src/libc.zig
```diff
@@ -0,0 +1,1074 @@
...
+    if (!elf_header.is_64) {
+        print.printError("ELF header from \"{s}\" seems to not be from a 64 bit binary", .{self_exe_path});
+        return error.ElfNot64Bit;
+    }
+
+    var dynamic_symbols_table_offset: u64 = 0;
+    var dynamic_symbols_table_size: u64 = 0;
+
+    try self_exe_file.seekTo(elf_header.shoff);
+    const section_headers = try std.heap.page_allocator.alloc(std.elf.Elf64_Shdr, elf_header.shnum);
```
jaronoff97: should we be passing the allocator down from the caller here? My zig knowledge is still growing, but I believe that's the idiom.

README.md
basti1302: Oh yes, absolutely. And guess what, they all already have the OTEL_INJECTOR_K8S_* name prefix in the code! These mappings have been implemented in the zig code in the previous version already, I just added docs for them here, but managed to screw up the env var names here in the docs. 🤦   

So, thanks for catching that 😃 👍 👍

src/libc.zig
basti1302: That's a good call, will change this.

README.md
basti1302: Fixed.

src/libc.zig
basti1302: Fixed. I think we actually should use the pattern of passing down the allocator in a couple more places (which have not been changed in this PR, so out of scope for here). I'll create a new tracking issue for that.

basti1302: -> https://github.com/open-telemetry/opentelemetry-injector/issues/178

README.md
```diff
@@ -145,21 +134,69 @@ Check the following for details about the auto-instrumtation agents and further
 - [Node.js](https://opentelemetry.io/docs/zero-code/js/configuration/)
 - [.NET](https://opentelemetry.io/docs/zero-code/dotnet/configuration/)
 
+### Environment Modifications
+
+Here is an overview of the modifications that the injector will apply:
+
+* It sets (or appends to) `NODE_OPTIONS` to activate the .
```
atoulme:
```suggestion
* It sets (or appends to) `NODE_OPTIONS` to activate the instrumentation agent.
```

README.md
```diff
@@ -145,21 +134,69 @@ Check the following for details about the auto-instrumtation agents and further
...
+  key-value pair takes priority.
+  The following environment variables and resource attributes are supported:
+    * `OTEL_INJECTOR_K8S_NAMESPACE_NAME` will be translated to `k8s.namespace.name`
+    * `OTEL_INJECTOR_K8S_POD_NAME` will be translated to `k8s.pod.name`
+    * `OTEL_INJECTOR_K8S_POD_UID` will be translated to `k8s.pod.uid`
+    * `OTEL_INJECTOR_K8S_CONTAINER_NAME` will be translated to `k8s.container.name`
+    * `OTEL_INJECTOR_SERVICE_NAME` will be translated to `service.name`
+    * `OTEL_INJECTOR_SERVICE_VERSION` will be translated to `service.version`
+    * `OTEL_INJECTOR_SERVICE_NAMESPACE` will be translated to `service.namespace`
+    * `OTEL_INJECTOR_RESOURCE_ATTRIBUTES` is expected to contain key-value pairs
```
atoulme: maybe lead with that env var? Is there a reason to use any other one?

README.md
```diff
@@ -145,21 +134,69 @@ Check the following for details about the auto-instrumtation agents and further
...
+Set the environment variable `OTEL_INJECTOR_LOG_LEVEL` to change the log level.
+Valid values are:
+- `debug`
+- `info`
+- `warn`
+- `error` - this is the default value
+- `none` - suppress all log output from the injector; this is useful for scenarios where you pipe `stderr` into another
+  process and parse it.
+
+The injector's log message will be written to `stderr` of the process the is being instrumented.
```
atoulme:
```suggestion
The injector's log message will be written to `stderr` of the process that is being instrumented.
```

README.md
```diff
@@ -212,15 +249,42 @@ include_paths=/utilities/*
...
+5. Again, if looking up any of the symbols fails, the injector aborts injection and hands back control to the host
+   process.
+6. Use the pointer to the `__environ` symbol to read the current environment of the process (before adding or modifying
+   any environment variables).
+7. Use the pointer to the `setenv` symbol to add or modify environment variables to add and activate OpenTelemetry
+   SDKs/auto-instrumentation agents for supported runtimes (e.g. `NODE_OPTIONS`, `JAVA_TOOL_OPTIONS`,
+   `CORECLR_PROFILER`).
+8. Use the pointer to the `setenv` symbol to add or modify `OTEL_RESOURCE_ATTRIBUTES`.
+
+There is a much more detailed explanation of this approach, and on alternative appraoches and the intricate design
```
atoulme:
```suggestion
There is a much more detailed explanation of this approach, and on alternative approaches and the intricate design
```

atoulme (approved):
LGTM - some comments on the README.md changes.

README.md
basti1302: This really comes from the Dash0 injector's Kubernetes focus and it's usage in a K8s operator.
Having e.g. OTEL_INJECTOR_SERVICE_NAME as a single env var enables mapping pod labels to resource attributes with a [field selector](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables), e.g. an operator using the injector can add

```
- name: OTEL_INJECTOR_SERVICE_NAME
   valueFrom:
     fieldRef:
       fieldPath: metadata.labels['app.kubernetes.io/name']
```

This is in line with K8s semconv recommendations: https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated

I'll swap the order as suggested and add something to explain what the intended usage of the individual env vars is 👍

PR #172, opened 2025-12-17
Deinit configuration

jack-berg:
Very new to zig, but I think #148 introduces a memory leak since `InjectorConfiguration` now includes a `HashMap` which needs to be freed when it goes out of scope.

https://github.com/open-telemetry/opentelemetry-injector/pull/148#discussion_r2627401235


jack-berg:
Nvm this approach is bad. The cached copy of `InjectorConfiguration` needs to live forever and calling `deinit` on it is causing the build to fail when subsequent calls to `getenv` try to reference the deinit-ed cached copy.

It may be worth finding a way to have the cached `InjectorConfiguration` use a slice of keyvalue pairs instead of a stringhashmap.

PR #174, opened 2025-12-17
fix(config): continue to read after overly long line

basti1302:
The update to Zig 0.15.2
(https://github.com/open-telemetry/opentelemetry-injector/pull/152) broke an edge case in the config parser: After encountering a overly long line, the config parser would stop reading the remaining lines of the config file, and instead of only ignoring the overly long line, would ignore the rest of the file.

While PR #152 tried to mitigate that by doubling the buffer size (and thus the problem only would occur on lines over 16k characters), it nevertheless only makes the bug more unlikely, but not impossible.

The better pattern to use here is to handle error.StreamTooLong inside the while-loop, so that the while-loop is not aborted by going into its else-branch.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -12,6 +12,7 @@ const testing = std.testing;
 
 const config_file_path = "/etc/opentelemetry/otelinject.conf";
 const max_line_length = 8192;
+const empty_string = @constCast("");
 
 const dotnet_path_key = "dotnet_auto_instrumentation_agent_path_prefix";
 const jvm_path_key = "jvm_auto_instrumentation_agent_path";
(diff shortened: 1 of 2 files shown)


src/config.zig
```diff
@@ -193,18 +194,9 @@ fn parseConfiguration(
...
     var reader = config_file.reader(&buf);
-
-    while (reader.interface.takeSentinel('\n')) |line| {
-        if (line.len > max_line_length) {
-            print.printError(
-                "A line in configuration file {s} exceeds the maximum allowed length of {d} characters and will be ignored",
-                .{ cfg_file_path, max_line_length },
-            );
-            continue;
-        }
```
basti1302, line 207: This is no longer needed, since lines with length > max_line_length are now handled in takeSentinelOrDiscardOverlyLongLine.

src/config.zig
```diff
@@ -230,6 +216,24 @@ fn parseConfiguration(
...
+                .{ cfg_file_path, max_line_length },
+            );
+            // Ignore lines that are too long for the buffer; advance the the read positon to the next delimiter to
+            // avoid stream corruption.
+            _ = try reader.interface.discardDelimiterInclusive('\n');
+            return empty_string;
+        },
+        else => |leftover_err| return leftover_err,
+    }
+}
```
basti1302, line 235: Note: The test "readConfigurationFile: does not parse overly long lines" would fail when extending the first long line to over 16k characters, this no longer happens now. (I ultimately didn't commit the change to the unit test asset because that 16k limit is no longer relevant now.)

src/config.zig
```diff
@@ -193,18 +194,9 @@ fn parseConfiguration(
     cfg_file_path: []const u8,
     comptime applyKeyValueToConfig: ConfigApplier,
 ) void {
-    const max_buffer_len = max_line_length * 2;
-    var buf: [max_buffer_len]u8 = undefined;
```
basti1302, line 197: With the fix, there is no need to provide an even bigger buffer.

basti1302 (commented):
cc @inge4pres I know we talked about this change quite a bit back in your PR. At the time I didn't really understand the consequences completely. While [bringing over](https://github.com/open-telemetry/opentelemetry-injector/pull/166) similar file parsing logic from the dash0-operator repo and updating it to 0.15.2, I noticed that this can be improved.

atoulme (approved):

PR #175, opened 2025-12-17
Fix index out of bounds error

### LINKED ISSUE #173: Bug in new release
### Component(s)

_No response_

### What happened?

I'm trying out the new release, and adding some content to the new `/etc/opentelemetry/default_auto_instrumentation_env.confg` file like so:

```
OTEL_EXPORTER_OTLP_ENDPOINT=<endpoint>
OTEL_EXPORTER_OTLP_HEADERS=Authorization=<auth_header>
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
```

After setting this, almost any process I start crashes with errors which look like:

```
thread 18798 panic: index out of bounds: index 15, len 14
aborting due to recursive panic
Aborted
```

Notably, I can't use vim to change the contents of `/etc/opentelemetry/default_auto_instrumentation_env.confg` because vim crashes!

```
jberg@berg-home:~/code/injector-playground/wildfly-app$ sudo vim /etc/opentelemetry/default_auto_instrumentation_env.conf 
thread 20409 panic: index out of bounds: index 15, len 14
aborting due to recursive panic
Aborted
```

Luckily, and for reasons I don't understand, I can forcibly delete the file and recreate it to restore the function of my system:
```
sudo rm -rf /etc/opentelemetry/default_auto_instrumentation_env.conf 
sudo touch /etc/opentelemetry/default_auto_instrumentation_env.conf
```

I could also remove the script from `/etc/ld.so.preload`... 

### Version

v0.0.2-20251216

jack-berg:
Fixes #173

### DIFF
--- a/.chloggen/env_vars_bug_fix.yaml
+++ b/.chloggen/env_vars_bug_fix.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: bug_Fix
+change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
 component: environment


jack-berg:
FYI I built distributions locally from this with `make rpm-package deb-package` and verified that the problem I describe in #173 goes away.

test/docker/default_auto_instrumentation_env.conf
```diff
@@ -1 +1,2 @@
 CUSTOM_ENV_VAR=custom-value-123
+ANOTHER_CUSTOM_ENV_VAR=custom-value-456
```
jack-berg, line 2: Build fails if you add this change, and don't have the corresponding fix in `root.zig`.

mmanciop: Yes, I tried the same repro today before I approved :-)

PR #180, opened 2025-12-23
chore: consistently pass allocator to functions that allocate memory

### LINKED ISSUE #178: Pass down allocator into functions that need to allocate memory
Right now, the code mostly references one globally declared page allocator (src/alloc.zig#page_allocator) directly, whenever it needs to allocate memory. We should move to a more idiomatic pattern, where calling code (e.g. src/root.zig) passes down an allocator into the functions requiring one.

That would also allow us to use std.testing.allocator to check for memory leaks.

basti1302:
Previously, the code mostly referenced one globally declared page allocator (src/alloc.zig#page_allocator) directly, whenever it needed to allocate memory.

This change moves to a more idiomatic pattern, where calling code (e.g. src/root.zig) passes down an allocator into the functions requiring one.

In production code, we now only declare one allocator at the start of initEnviron. This is passed to all other functions that need an allocator. This also makes it much more obvious which functions allocate memory and which functions are allocation free.

Ultimately, all unit tests should probably use std.testing.allocator and free everything they allocated, this is not in scope for this change (a lot of test code still uses std.heap.page_allocator directly).

Fixes #178.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -3,7 +3,6 @@
 
 const std = @import("std");
 
-const alloc = @import("allocator.zig");
 const print = @import("print.zig");
 const test_util = @import("test_util.zig");
 const patterns_util = @import("patterns_util.zig");
@@ -47,9 +46,17 @@ pub const InjectorConfiguration = struct {
     exclude_paths: [][]const u8,
     include_args: [][]const u8,
     exclude_args: [][]const u8,
+
+    pub fn freeAll(self: *InjectorConfiguration, allocator: std.mem.Allocator) void {
+        allocator.free(self.all_auto_instrumentation_agents_env_path);
+        self.all_auto_instrumentation_agents_env_vars.deinit();
+        allocator.free(self.dotnet_auto_instrumentation_agent_path_prefix);
+        allocator.free(self.jvm_auto_instrumentation_agent_path);
+        allocator.free(self.nodejs_auto_instrumentation_agent_path);
+    }
 };
 
-const ConfigApplier = fn (key: []const u8, value: []u8, file_path: []const u8, configuration: *InjectorConfiguration) void;
+const ConfigApplier = fn (gpa: std.mem.Allocator, key: []const u8, value: []u8, file_path: []const u8, configuration: *InjectorConfiguration) void;
 
 const default_dotnet_auto_instrumentation_agent_path_prefix = "/__otel_auto_instrumentation/dotnet";
 const default_jvm_auto_instrumentation_agent_path = "/__otel_auto_instrumentation/jvm/opentelemetry-javaagent.jar";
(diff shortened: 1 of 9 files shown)


src/config.zig
```diff
@@ -47,9 +46,17 @@ pub const InjectorConfiguration = struct {
     exclude_paths: [][]const u8,
     include_args: [][]const u8,
     exclude_args: [][]const u8,
+
+    pub fn freeAll(self: *InjectorConfiguration, allocator: std.mem.Allocator) void {
```
jaronoff97, line 50: would it make sense to use an arena allocator here?

src/config.zig
```diff
@@ -165,16 +172,22 @@ fn readConfigurationFile(cfg_file_path: []const u8, configuration: *InjectorConf
...
+        gpa,
+        configuration,
+        config_file,
+        cfg_file_path,
+        applyKeyValueToGeneralOptions,
+    );
 }
 
-fn applyKeyValueToAllAgentsEnv(key: []const u8, value: []u8, _file_path: []const u8, _configuration: *InjectorConfiguration) void {
+fn applyKeyValueToAllAgentsEnv(_: std.mem.Allocator, key: []const u8, value: []u8, _file_path: []const u8, _configuration: *InjectorConfiguration) void {
```
jaronoff97, line 184: if this doesn't need the param, why add it?

src/config.zig
```diff
@@ -235,9 +249,11 @@ fn takeSentinelOrDiscardOverlyLongLine(reader: *std.fs.File.Reader, cfg_file_pat
 }
 
 test "readConfigurationFile: file does not exist" {
-    var configuration = createDefaultConfiguration();
+    const allocator = std.heap.page_allocator;
```
jaronoff97, line 252: shouldn't tests should use the test allocator `std.testing.allocator`?

src/config.zig
basti1302: Yes, absolutely. This is part of https://github.com/open-telemetry/opentelemetry-injector/pull/184. I wanted to keep the PRs somewhat manageable after the last very large PR.

src/config.zig
basti1302: ~That is a good question. I'll remove it.~

basti1302: Ah, no, now I remember. Both `applyKeyValueToGeneralOptions` and `applyKeyValueToAllAgentsEnv` need to have the same signature, they must both adhere to the type `ConfigApplier`, because they are both passed to `parseConfiguration`. 

So that's the reason for the unused parameter.

src/config.zig
basti1302: Yes, and that change is also part of PR #184. Changing the allocator to std.testing.allocator already in this PR would make the test fail (because there are still allocations that are not freed).

src/config.zig
jaronoff97: ah np! I hadn't seen that until now. ✔️

PR #183, opened 2025-12-23
chore: use sentinel-terminated slices internally

### LINKED ISSUE #181: Consistently use std.testing.allocator in all unit tests, free memory allocated in unit tests
### Component(s)

_No response_

### Describe the issue you're reporting

Follow up to #178 / #180.

### Tip

basti1302:
Use sentinel-terminated slices
(https://ziglang.org/documentation/0.15.2/#Sentinel-Terminated-Slices)
internally for all strings, instead of the more clumsy
sentinel-terminated many-pointers
(https://ziglang.org/documentation/0.15.2/#Sentinel-Terminated-Pointers).

Convert the sentinel-terminated slices to sentinel-terminated
many-pointers at the last possible moment, when calling setenv.
We can also get rid of the explicit @ptrCast(xxx.ptr), since
sentinel-terminated slices can be coerced to sentinel-terminated
many-pointers.

This allows to more easily free the produced slices in unit tests, which
in turn enables using std.testing.allocator, which in turn helps to uncover
memory leaks.

As a result, this commit also adds a couple of allocator.free calls to
production code to fix some minor leaks that were uncovered because of
this (e.g. preliminary strings that are later copied over to another
memory region and can then be discarded).

This is most of the work required to fix #181, but two usage patterns
with std.heap.page_allocator in unit tests remain for now, in
config.zig, and in resource_attributes_test.zig.

### DIFF
--- a/src/dotnet.zig
+++ b/src/dotnet.zig
@@ -13,14 +13,26 @@ const test_util = @import("test_util.zig");
 const testing = std.testing;
 
 pub const DotnetValues = struct {
-    coreclr_enable_profiling: types.NullTerminatedString,
-    coreclr_profiler: types.NullTerminatedString,
-    coreclr_profiler_path: types.NullTerminatedString,
-    additional_deps: types.NullTerminatedString,
-    shared_store: types.NullTerminatedString,
-    startup_hooks: types.NullTerminatedString,
-    otel_auto_home: types.NullTerminatedString,
+    coreclr_enable_profiling: [:0]const u8,
+    coreclr_profiler: [:0]const u8,
+    coreclr_profiler_path: [:0]u8,
+    additional_deps: [:0]u8,
+    shared_store: [:0]u8,
+    startup_hooks: [:0]u8,
+    otel_auto_home: [:0]u8,
+
+    pub fn freeAll(self: DotnetValues, allocator: std.mem.Allocator) void {
+        allocator.free(self.coreclr_profiler_path);
+        allocator.free(self.additional_deps);
+        allocator.free(self.shared_store);
+        allocator.free(self.startup_hooks);
+        allocator.free(self.otel_auto_home);
+    }
 };
+
+const coreclr_enable_profiling_value = "1";
+// See https://opentelemetry.io/docs/zero-code/dotnet/configuration/#net-clr-profiler.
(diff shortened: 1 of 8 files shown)


src/dotnet.zig
```diff
@@ -13,15 +13,26 @@ const test_util = @import("test_util.zig");
...
+        allocator.free(self.coreclr_profiler_path);
+        allocator.free(self.additional_deps);
+        allocator.free(self.shared_store);
+        allocator.free(self.startup_hooks);
+        allocator.free(self.otel_auto_home);
+    }
 };
 
+const coreclr_enable_profiling_value = "1";
+const coreclr_profiler_value = "{918728DD-259F-4A6A-AC2B-B85E1B658318}";
```
jaronoff97, line 35: i know this was here before, but what is this value / where is it derived from?

src/jvm.zig
```diff
@@ -57,95 +61,114 @@ fn doCheckOTelJavaAgentJarAndGetModifiedJavaToolOptionsValue(
...
         );
-    try test_util.expectWithMessage(modifiedJavaToolOptions == null, "modifiedJavaToolOptions == null");
+    try test_util.expectWithMessage(modified_java_tool_options == null, "modified_java_tool_options == null");
 }
 
 test "doCheckOTelJavaAgentJarAndGetModifiedJavaToolOptionsValue: should return null if the Java agent cannot be accessed (original value present)" {
     const path = try std.fmt.allocPrint(testing.allocator, "/invalid/path", .{});
     defer testing.allocator.free(path);
-    const modifiedJavaToolOptions =
+    const modified_java_tool_options =
```
jaronoff97, line 76: why did this change case?

src/dotnet.zig
basti1302: Each .NET profiler needs to have a unique ID. This is the ID of the OTel .NET/CLR instrumentation. It comes from here: https://opentelemetry.io/docs/zero-code/dotnet/configuration/#net-clr-profiler.

But it's a good question, I'll add a comment with that link.

src/jvm.zig
basti1302: This was not in line with the official style guide, i.e. https://ziglang.org/documentation/0.15.2/#toc-Names -> 

> camelCaseFunctionName, TitleCaseTypeName, snake_case_variable_name

I just aligned it with how it is supposed to be written.

PR #185, opened 2025-12-29
Limit all_auto_instrumentation_agents_env_path env vars to entries prefixed with OTEL_

### LINKED ISSUE #176: Limit `all_auto_instrumentation_agents_env_path` to only load `OTEL_` env vars
### Component(s)

_No response_

The power of the LD_PRELOAD approach the injector takes means we need additional diligence:

- Code quality is especially important, since bugs like #173 can cause all (or most) processes to crash and cause system instability
- The injector becomes a natural attack vector for those looking to exploit vulnerabilities.

With respect to the second point, I think we need to add some constraints to the recently added capability to inject arbitrary environment variables in #148.

Limit env var inection with `all_auto_instrumentation_agents_env_path` to only support keys starting with `OTEL_*`.

I also considered limiting to an allow list of the well known env vars described in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md). But I think this is ultimately not going to be enough, as many languages have important language-specific env vars not part of the spec. See [otel java agent docs](https://opentelemetry.io/docs/zero-code/java/agent/configuration/ for some examples. 

Limiting to those starting with `OTEL_` seems to strike the right balance: 

- its easy to describe / document
- it won't require maintaining a constantly updating allow list
- it limits the blast radius of the injector to make it less attractive as an attack vector 

### Additional context

_No response_

### Tip

jack-berg:
Resolves #176

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -3,7 +3,6 @@
 
 const std = @import("std");
 
-const alloc = @import("allocator.zig");
 const print = @import("print.zig");
 const test_util = @import("test_util.zig");
 const patterns_util = @import("patterns_util.zig");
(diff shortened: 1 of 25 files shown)


src/config.zig
```diff
@@ -169,6 +170,11 @@ fn readConfigurationFile(cfg_file_path: []const u8, configuration: *InjectorConf
 }
 
 fn applyKeyValueToAllAgentsEnv(key: []const u8, value: []u8, _file_path: []const u8, _configuration: *InjectorConfiguration) void {
+    if (!std.mem.startsWith(u8, key, otel_env_var_prefix)) {
+        print.printError("env var {s} does not start with {s}. ignoring.", .{ key, otel_env_var_prefix });
```
basti1302: - Should this maybe be a warning instead?
- Can you please not abbreviate env var in a user facing error message? I think we use "environment variable" in other messages.

```suggestion
        print.printWarn("environment variable {s} does not start with {s}, ignoring.", .{ key, otel_env_var_prefix });
```

test/docker/default_auto_instrumentation_env.conf
```diff
@@ -1,2 +1,3 @@
+OTEL_SDK_DISABLED=true
 CUSTOM_ENV_VAR=custom-value-123
 ANOTHER_CUSTOM_ENV_VAR=custom-value-456
```
basti1302, line 4: IIRC we added a second environment variable here for a specific regression. Not sure if this is still relevant, but maybe we should have two OTEL_ names in here?

jack-berg: Yup good call #175 fixes an off by one error which only manifests with multiple env variables. With this new code filtering which environment variables being set, we need to make sure the size of the filtered set is > 1 to exercise this code.

basti1302 (approved):
LGTM

PR #186, opened 2025-12-30
build: only rebuild dist if source files have changed

basti1302:
Use make's mtime checks to determine whether "make dist" actually needs to rebuild the executable.


basti1302:
This will come in handy in https://github.com/open-telemetry/opentelemetry-injector/pull/155, we can build the binary once for each arch and then just re-use it for the package-integration tests or for building the actual deb/rpm packages.

PR #190, opened 2026-01-09
ci: fix tag trigger in Github action workflow build

basti1302:
The expression is interpreted as a glob, not a regex; thus the previous expression ".*" would only match tags starting with a literal dot.

The new expression should match the tag format mentioned in RELEASING.md.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     tags:
-      - ".*"
+      - "v[0-9]+.[0-9]+.[0-9]+**"
   pull_request:
 
 permissions:
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -6,7 +6,7 @@ on:
     inputs:
       candidate:
         required: true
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     tags:
-      - ".*"
+      - "v[0-9]+.[0-9]+.[0-9]+**"
```
mmanciop, line 8: Looks fishy, why the double asterisk?

mmanciop: Because glob

basti1302: The part `v[0-9]+.[0-9]+.[0-9]` matches regular releases, the `**` part matches an optional qualifier like `-20260109` for pre-releases. With a regex you might write this as `(-.*)?` but it ain't a regex. 

See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet.

basti1302: Actually, a single `*` would have also worked, the difference being that the single asterisk does not match `/` (which we don't need I guess). Whatever.

PR #198, opened 2026-01-12
 test(integration): add regression test for binaries without libc

basti1302:
This test case verifies that the injector does not crash executables
that are subject to dynamic linking but do _not_ bind a libc


basti1302:
Stacked PR, #197 needs to be merged first.

grcevski (approved):
LGTM!

PR #201, opened 2026-01-14
chore: align in-code default paths for agents w. defaults from packages

basti1302:
The previous default paths did not really make much sense, they stem historically from the fact that this was a port of the Dash0 injector which is only used in K8s via an init container, and there the goal was to use paths that do not collide with any existing file system structure.

Since we currently only provide RPM and DEB packages, and both have the same default config, we should use the default paths from packaging/fpm/etc/opentelemetry/otelinject.conf for the in-code defaults as well.


grcevski (commented):
This makes sense to me, it standardises the way the paths are defined, and I liked the use of the term `jvm` instead of `java`.

mmanciop (approved):

mmanciop (commented):
Test need fixing, but overall LGTM

src/config.zig
```diff
@@ -249,7 +249,7 @@ test "readConfigurationFromPath: all configuration values from file, no environm
         configuration.jvm_auto_instrumentation_agent_path,
     );
     try testing.expectEqualStrings(
-        "/custom/path/to/node_js/node_modules/@opentelemetry-js/otel/instrument",
+        "/custom/path/to/js/node_modules/@opentelemetry/auto-instrumentations-node/build/src/register.js",
```
jaronoff97: in operator land this is `nodejs`, does it make sense to keep that consistent here?

jaronoff97 (approved):
just one thought, otherwise fine.

src/config.zig
basti1302: The [existing default](https://github.com/open-telemetry/opentelemetry-injector/blob/v0.0.3/packaging/fpm/etc/opentelemetry/otelinject.conf#L6) used `/js/`, so I went with this. But I also like `/nodejs/` better, since we certainly don't instrument client-side JS with the LD_PRELOAD injector :-) 

So `nodejs` it is.

PR #204, opened 2026-01-16
build: make package build logs less noisy

basti1302:
Also: fix construction of paths with double slashes (e.g. /tmp/tmp.sElUvhJcuS//usr/lib/opentelemetry/...) in package build scripts.

### DIFF
--- a/packaging/fpm/common.sh
+++ b/packaging/fpm/common.sh
@@ -70,10 +70,11 @@ download_nodejs_agent() {
     pushd "$(dirname "$dest")"
     mkdir -p "nodejs"
     pushd "nodejs"
-    npm pack "@opentelemetry/auto-instrumentations-node@${tag#v}"
-    mv ./*.tgz otel-js.tgz
-    npm install --global=false otel-js.tgz
-    rm otel-js.tgz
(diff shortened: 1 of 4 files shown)


packaging/fpm/common.sh
```diff
@@ -128,24 +129,24 @@ setup_files_and_permissions() {
     nodejs_agent_release="$(tail -n 1 <"$NODEJS_AGENT_RELEASE_PATH")"
     dotnet_agent_release="$(tail -n 1 <"$DOTNET_AGENT_RELEASE_PATH")"
 
-    mkdir -p "$buildroot/$(dirname $libotelinject_INSTALL_PATH)"
```
basti1302, line 131: $(dirname $libotelinject_INSTALL_PATH) and basically all other variables used below in a similar manner denote absolute paths, like `/usr/lib/opentelemetry/something`, hence adding a literal `/` character between buildpath and $XXX_PATH would result in `/tmp/tmp.sElUvhJcuS//usr/lib/opentelemetry/something` (note the double slash).

This was inconsequential, bash apparently just ignores the double slash, but looked odd in the build log.

atoulme (approved):

PR #210, opened 2026-01-21
chore: init pid for log messages only once

basti1302:
Commit 8f1535a9e19d922e3c920699098a4e44098a2bf1 introduced a change that crashes the Zig unit tests when executed directly on MacOS - in particular the std.os.linux.getpid() crashes when the OS is not Linux.

While support for any OS besides Linux is currently not a goal of this project, enabling developers to run the Zig unit tests directly on their native OS has value.

This commit also changes the initialization in src/print.zig to acquire the pid once, and then reuse that value when printing log messages.

### DIFF
--- a/src/print.zig
+++ b/src/print.zig
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+const builtin = @import("builtin");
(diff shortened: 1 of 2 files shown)


mmanciop (approved):

grcevski (commented):
Nice, thanks for fixing this.

PR #216, opened 2026-01-22
feat: add opt-in support for Python auto-instrumentation

basti1302:
For now, auto-instrumentation for Python is off by default and no
default Python auto-instrumentation agent is included in the system
packages.

Python instrumentation needs to be enabled explicitly by providing a
Python auto-instrumentation agent and a custom otelinject.conf file with
python_auto_instrumentation_agent_path_prefix set or by setting
PYTHON_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX.

Be aware that it is not always safe to inject the existing zero-code
instrumentation for Python due to possible dependency conflicts
(protobuf etc.)


grcevski:
This is great, it opens up the path for adding maybe some scanning capabilities with other tools for possible problem libraries and having orchestrators decide if they should include Python.

PR #238, opened 2026-02-03
[chore] fix release regular expression

atoulme:
Two problems:
* `?:` bash doesn't support non-capturing groups. Drop it.
* `.*` replaced by `\w+`: bash is applying greediness to the first `.*` it meets - reducing to word characters helps. There is no way to apply ungreediness to RE regexps.

### DIFF
--- a/.github/workflows/scripts/create-tag-for-release.sh
+++ b/.github/workflows/scripts/create-tag-for-release.sh
@@ -11,7 +11,7 @@ cd "$(dirname "${BASH_SOURCE[0]}")/../../.."
 COMMIT_MSG=$(git log -1 --pretty=%B)
 
 # Check if commit message matches release pattern and extract version
-if [[ "$COMMIT_MSG" =~ ^docs:\ update\ changelog\ to\ prepare\ release\ (v[0-9]+\.[0-9]+\.[0-9]+(-?\w+)?)\ .*$ ]]; then
+if [[ "$COMMIT_MSG" =~ ^docs:\ update\ changelog\ to\ prepare\ release\ (v[0-9]+\.[0-9]+\.[0-9]+(-[[:alnum:]]+)?).*$ ]]; then
   VERSION="${BASH_REMATCH[1]}"
   echo "Found release commit for version: $VERSION."
 


.github/workflows/scripts/create-tag-for-release.sh
```diff
@@ -11,7 +11,7 @@ cd "$(dirname "${BASH_SOURCE[0]}")/../../.."
 COMMIT_MSG=$(git log -1 --pretty=%B)
 
 # Check if commit message matches release pattern and extract version
-if [[ "$COMMIT_MSG" =~ ^docs:\ update\ changelog\ to\ prepare\ release\ (v[0-9]+\.[0-9]+\.[0-9]+(?:-?.*)?)\ .*$ ]]; then
+if [[ "$COMMIT_MSG" =~ ^docs:\ update\ changelog\ to\ prepare\ release\ (v[0-9]+\.[0-9]+\.[0-9]+(-?\w+)?)\ .*$ ]]; then
```
basti1302: I tested this locally and at least in my version of bash it does not work with this commit message:

```
docs: update changelog to prepare release v0.0.4-20260202 (#230)

Co-authored-by: opentelemetrybot
<107717825+opentelemetrybot@users.noreply.github.com>
```

Results in:
```
The most recent commit does not seem to be a release commit.
```

I think character classes like `\w` are also not supported in bash?


Here is another suggestion that seems to work for both cases:

```suggestion
if [[ "$COMMIT_MSG" =~ ^docs:\ update\ changelog\ to\ prepare\ release\ (v[0-9]+\.[0-9]+\.[0-9]+(-[[:alnum:]]+)?).*$ ]]; then
```

I have tested this version of the regex 👆 against both commit messages 

```
docs: update changelog to prepare release v0.0.4-20260202 (#230)

Co-authored-by: opentelemetrybot
<107717825+opentelemetrybot@users.noreply.github.com>
```
➡️
```
Found release commit for version: v0.0.4-20260202
```

and

```
commit message: docs: update changelog to prepare release v0.0.4 (#234)

Co-authored-by: opentelemetrybot
<107717825+opentelemetrybot@users.noreply.github.com>
```
➡️
```
Found release commit for version: v0.0.4
```

PR #245, opened 2026-02-03
chore: minor doc fixes

kerneljack:
I haven't managed to do much more than read the docs so far and find some typos, so I thought I'll submit a small fix.

I don't think I need a changelog entry for this, but am happy to add one if needed, please let me know.


README.md
kerneljack, line 52: a question about this path - if I extract the .deb file, the actual path and the value in the `/etc/opentelemetry/otelinject.conf` file doesn't have the `jvm` directory in it. So I think the path needs to be corrected.

atoulme: please open a separate PR

PR #251, opened 2026-02-09
docs(design): add section that explains why we are using Zig

basti1302:


basti1302:
I've been asked this a couple of times since we started, most recently at OTel Unplugged. Rust is brought up a lot as an alternative. I think it makes sense to add some motivation for using Zig, in particular since it is still a somewhat niche language.

DESIGN.md
```diff
@@ -265,3 +265,31 @@ This is because the binary is dynamically linked, hence it is affected by `LD_PR
...
+  Not relying on garbage collection also works in favor of avoiding accidental CPU usage overhead.
+
+Zig has combination of characteristics that make it a good match for the items listed above:
+* It does not bind a libc flavor, but still provides a rich standard library.
+* It does not have automatic memory management, but relies on manual memory management instead.
+
+Let's compare this to some other programming languages that might be used for the injector:
+* The preference for manual memory management rules out Go.
+* Rust is also brought up fairly often, but it binds libc by default.
+  `#![no_std]` is available, but that also removes access to basic functionality (printing, working with strings).
```
basti1302, line 291: I _think_ this is true, but I do not really have any Rust experience to speak of, if some of you know their way around Rust, please let me know whether this is accurate.

DESIGN.md
```diff
@@ -265,3 +265,31 @@ This is because the binary is dynamically linked, hence it is affected by `LD_PR
...
+* Traditionally, similar technology has been written in C++ or plain C.
+  This is possible, but again, without binding to libc, you would have to rewrite a lot of basic functionality.
+  Also, Zig avoids a lot of the footguns that C code can introduce, that is, code written in Zig is probably usually
+  safer than C.
```
grcevski, line 295: I would add one more, Zig compilation time is 10x faster compared to Rust, it improves developer happiness.

basti1302: Oh, interesting, I didn't know. Do you have a source or is this more from personal experience :-)

(I don't have sources for any of the other things, so I'm fine adding it regardless, this is mostly me being curious.)

grcevski: Here's one article I read last fall https://mitchellh.com/writing/zig-builds-getting-faster, taking out LLVM out of the picture is the big win. Obviously it's hard to compare since not many real projects have implementations in both languages. I think the caching that zig does locally helps a ton, here's a comparison of clang/llmv, zig and tiny_cc https://forum.nim-lang.org/t/8677. In my experience clang (for C++) and rustc have similar performance, assuming cargo caches all you need for the build, like zig does.

Maybe it's not worth mentioning :), it will be generally faster, but I think it would also depend on the project. Maybe for the injector itself it won't be much of a difference. Zig is much simpler language, so it will be always faster to compile than C++ or Rust.

basti1302: I added a bullet item for this.

mmanciop: I'd not mention it. It's flame-bait for some folks, without being germane to the decision of using Zig. I would have used anything that gave me a decent SDK, no dependency on Libc, and ease of handling Elf (and the last one, I might have done without and done my own lib)

basti1302: Michele, your last comment came in after I merged, so the bullet item is there now, albeit phrased fairly neutral and without a direct comparison:

>  Zig also supposedly compiles fairly [fast](https://mitchellh.com/writing/zig-builds-getting-faster), making it
  pleasant to work with (that said, the injector code base is quite small, so compiler perfomance is not very crucial).

We can remove it again, I do not have a strong opinion either way. Actually, not even a weak opinion. :)

mmanciop: Not worth kicking it out again. I had actually posted it in time, but GitHub was wobbly and the comment stayed in my browser. It apparently got sent after a page reload or so :-)

Issue #240, opened 2026-02-03
Add environment variable to disable all auto-instrumentation at once


basti1302:
### Is your feature request related to a problem? Please describe.

There is no simple way to tell the injector to not add auto-instrumentation.

### Describe the solution you'd like

OTEL_INJECTOR_DISABLED=true should disable the injector entirely.

This is useful in scenarios where the injector is added to all processes (e.g. all processes on a host, or all workloads in a Kubernetes cluster/namespace etc.) and you want to disable it for individual workloads.

### Describe alternatives you've considered

Currently, there would be two ways to achieve this:
1. Set the individual instrumentation env vars to the empty string, e.g. DOTNET_AUTO_INSTRUMENTATION_AGENT_PATH_PREFIX="" JVM_AUTO_INSTRUMENTATION_AGENT_PATH=""  NODEJS_AUTO_INSTRUMENTATION_AGENT_PATH="" respectively. This is cumbersome and it is also not future-proof (if the injector adds support for another runtime, you would have to add another env var)
2. Set LD_PRELOAD to the empty string. Depending how the injector has been added to the system (/etc/ld.so.preload, setting LD_PRELOAD globally) this might also be awkward to achieve.

Having a simple global off switch would be better than the workarounds listed above.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #256, opened 2026-02-11
Add option to completely disable the injector with an environment variable

grcevski:
I'm implementing option 1 in the related issue #240.

I had two approaches in mind for achieving this:
1. Short circuit all configuration parsing and environment variable evaluation by checking the option early.
2. Check the new environment variable in the environment parsing code and set the boolean for disabled there.

I opted in for option 1 for a few reasons:
1. It's faster to bail early.
2. If we do it late when the environment variables are parsed, the debug logs look strange. We say all the things we've set just to nullify it all at the end.

Relates to #240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ so/
 dist/
 
 .tools/
-.vscode/settings.json
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+    "zig.testArgs": [
+        "test",
+        "-lc",
+        "--test-filter",
+        "${filter}",
+        "${path}"
+    ]
+}


jaronoff97 (approved):

src/config.zig
```diff
@@ -154,6 +169,132 @@ fn readConfigurationFromPath(allocator: std.mem.Allocator, cfg_file_path: []cons
...
+        configuration.jvm_auto_instrumentation_agent_path,
+    );
+}
+
+test "readConfiguration: should return empty config when OTEL_INJECTOR_DISABLED is set to 'true' values" {
+    const allocator = testing.allocator;
+
+    const true_values = [_][]const u8{ "true", "TRUE", "True", "1", "t", "T" };
+
+    for (true_values) |value| {
```
basti1302, line 230: _praise:_ I like this test pattern! It feels pretty much like table driven tests from Go or similar. Less clutter and less repetition. 👍

.gitignore
```diff
@@ -5,3 +5,4 @@ so/
 dist/
 
 .tools/
+.vscode/settings.json
```
atoulme: is this related?

grcevski: I wasn't able to run the new tests I wrote and debug them in VSCode because we now check for libc not being linked. So  VSCode needs a modified run.args for the tests, like below:
```
{
    "zig.testArgs": [
        "test",
        "-lc",
        "--test-filter",
        "${filter}",
        "${path}"
    ]
}
```
I didn't want to pollute the repo with VSCode files if people don't use the tool, but adding a separate PR for just modifying the gitignore seemed like an overkill.

If folks think it's useful to have the vscode settings for the Zig extension, I can add this change. WDYT?

basti1302: I don't have a strong opinion either way. I don't use VSCode myself, but if adding the settings file reduces friction for what is probably one of the most popular IDEs these days, I slightly lean towards adding it. For folks that don't use VSCode, the file will simply be ignored, so there is no real downside to adding it right?

atoulme: feel free to add it, given that zig might be less well known to developers, we can be more explicit in guidance and enforcement of how folks should interface with the repository.

grcevski: OK, cool! I'll add it. To be honest the Zig VSCode extension did pretty good job on the source immediately as I started working on this project. I just needed this small tweak when we added the nolibc checking. With this you can click in the UI and run individual tests.

grcevski: Done.

.vscode/settings.json
mmanciop, line 1: Should not be committed

mmanciop (approved):
Minor nit. I like the testing :-)

.vscode/settings.json
grcevski: We discussed it [here](https://github.com/open-telemetry/opentelemetry-injector/pull/256#discussion_r2799770818) and Basti and Antoine said I should :). My initial commit was to add the ignore.

mmanciop: mmm okay then

PR #269, opened 2026-02-25
feat(config): add support for auto_instrumentation_disabled list

### LINKED ISSUE #240: Add environment variable to disable all auto-instrumentation at once

basti1302:
This adds a new supported key to the configuration file, named auto_instrumentation_disabled

The value can be either the literal "*" character to disable auto-instrumentation for all supported runtimes, or it can be a comma-separated list of runtimes, to selectively disable auto-instrumentation for these runtimes only.

In both cases, adding resource attributes via the injector is still enabled.

Also adds support for an environment variable with the same semantics, OTEL_INJECTOR_AUTO_INSTRUMENTATION_DISABLED.

fixes #240

### DIFF
--- a/.chloggen/add_option_to_disable_auto_instrumentation.yaml
+++ b/.chloggen/add_option_to_disable_auto_instrumentation.yaml
@@ -1,0 +1,11 @@
+change_type: enhancement
+component: injection
+note: Add option to disable auto-instrumentation for specific runtimes or all runtimes
+issues: [240]
+subtext: When the config file option `auto_instrumentation_disabled` or the environment variable
+  `OTEL_INJECTOR_AUTO_INSTRUMENTATION_DISABLED` is set to `*`, auto-instrumentation for all runtimes will be disabled.
+  When set to a comma-separated list of runtimes, auto-instrumentation for the listed runtimes will be disabled.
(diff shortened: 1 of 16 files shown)


grcevski (approved):
LGTM! Looks really good, I like the level of testing done on the new option.

PR #239, opened 2026-02-03
Packaging OTEP 4793

mmanciop:
Implementing the packaging described in [OTEP 4793](https://github.com/open-telemetry/opentelemetry-specification/pull/4793)

PR #291, opened 2026-03-17
change config file to manage composition

atoulme:
BREAKING CHANGE: To facilitate the composability of configuration needed in system packages, the configuration now moves under `/etc/opentelemetry/injector/`, with the main configuration file a `/etc/opentelemetry/injector/injector.conf`, merging with drop-in files under `/etc/opentelemetry/injector/conf.d/`.

The environment variable configuration file also moves from `/etc/opentelemetry/default_auto_instrumentation_env.conf` to `/etc/opentelemetry/injector/default_env.conf`.

The default auto-instrumentation agent paths are now empty. Paths are provided by per-language conf.d drop-in files installed by the respective language packages (e.g., `opentelemetry-java-autoinstrumentation` installs `jvm.conf`).

Users building custom container images must now explicitly configure the agent paths via conf.d files, the main configuration file, or environment variables.

This change has been extracted from #239.


src/config.zig
```diff
@@ -9,12 +9,18 @@ const patterns_util = @import("patterns_util.zig");
 
 const testing = std.testing;
 
-const default_config_file_path = "/etc/opentelemetry/otelinject.conf";
+const default_config_file_path = "/etc/opentelemetry/injector/otelinject.conf";
```
atoulme: moving under a folder helps so we can have a `conf.d` subfolder.

basti1302: Just to bring this up: This is a breaking change.

* Users of the system packages that have a custom/modified config file in `/etc/opentelemetry/otelinject.conf` will no longer get their custom settings, but the defaults.
* Users that use the binary directly, for example by building container images and putting a config file into the default location need to update their Dockerfiles. This would for example be the case for the Dash0 operator.

I think making this breaking change is fine, but it needs to be called out in the changelog very clearly.

src/config.zig
```diff
@@ -9,12 +9,18 @@ const patterns_util = @import("patterns_util.zig");
...
 const config_file_path_env_var = "OTEL_INJECTOR_CONFIG_FILE";
 const max_line_length = 8192;
 const empty_string = @constCast("");
 const otel_env_var_prefix = "OTEL_";
 
+// Agent paths are empty by default; they are enabled by installing conf.d drop-in files from the
+// respective language packages (e.g., opentelemetry-java-autoinstrumentation installs java.conf).
+const default_all_auto_instrumentation_agents_env_path = "/etc/opentelemetry/injector/default_env.conf";
```
basti1302, line 21: Same breaking change here, for users that have an existing file at  the old location, `/etc/opentelemetry/default_auto_instrumentation_env.conf`, this also needs to be called out in the changelog.

src/config.zig
```diff
@@ -81,15 +87,14 @@ pub const InjectorConfiguration = struct {
 
 const ConfigApplier = fn (gpa: std.mem.Allocator, key: []const u8, value: []u8, file_path: []const u8, configuration: *InjectorConfiguration) void;
 
-const default_dotnet_auto_instrumentation_agent_path_prefix = "/usr/lib/opentelemetry/dotnet";
-const default_jvm_auto_instrumentation_agent_path = "/usr/lib/opentelemetry/jvm/javaagent.jar";
-const default_nodejs_auto_instrumentation_agent_path = "/usr/lib/opentelemetry/nodejs/node_modules/@opentelemetry/auto-instrumentations-node/build/src/register.js";
+const default_dotnet_auto_instrumentation_agent_path_prefix = "";
+const default_jvm_auto_instrumentation_agent_path = "";
+const default_nodejs_auto_instrumentation_agent_path = "";
```
basti1302, line 93: Making the default empty is also a breaking change for users that do not use the system packages (where we would set defaults in the config I guess?), but e.g. build container images and just put the agents into these default locations. They now need to set the paths explicitly on their end.

What is the motivation of not providing any default paths in-code? I think this would not affect the system packages, since they probably always set defaults then in the config file(s).

src/config.zig
mmanciop: I would prefer us not to break stuff already. I'd rather go the composable path only when `/etc/opentelemetry/otelinject.conf` is not present.

mmanciop (dismissed):
I would prefer to avoid breaking changes like this, which likely will go unnoticed by users and will just result in apps not instrumenting (so, almost a silent failure)

.chloggen/config.yaml
```diff
@@ -1,25 +1,32 @@
-# The directory that stores individual changelog entries.
```
atoulme, line 1: revert those changes! the config.yaml file should not be changed.

src/config.zig
```diff
@@ -549,6 +555,57 @@ fn readConfigurationFile(arena_allocator: std.mem.Allocator, cfg_file_path: []co
     print.printDebug("successfully read configuration file from {s}.", .{cfg_file_path});
 }
 
+/// Reads configuration drop-in files from the conf.d directory. Each installed language package
+/// (e.g., opentelemetry-java-autoinstrumentation) places its configuration in this directory.
+/// Files are read in alphabetical order; later files can override earlier ones.
+fn readConfigurationDirectory(arena_allocator: std.mem.Allocator, configuration: *InjectorConfiguration) void {
```
atoulme, line 590: note: adding a test for this code, reuse unit test assets, make assertions on resulting configuration.

src/config.zig
mmanciop: I retract this objection :-)

src/config.zig
mmanciop: Done in 29b93a8

src/config.zig
mmanciop: Agreed in the SIG to this being a breaking change

src/config.zig
mmanciop: Agreed in the SIG to this being a breaking change

src/config.zig
mmanciop: Agreed in the SIG to this being a breaking change

.chloggen/config.yaml
mmanciop: Done

src/config.zig
```diff
@@ -9,12 +9,18 @@ const patterns_util = @import("patterns_util.zig");
...
 const testing = std.testing;
 
-const default_config_file_path = "/etc/opentelemetry/otelinject.conf";
+const default_config_file_path = "/etc/opentelemetry/injector/otelinject.conf";
 const config_file_path_env_var = "OTEL_INJECTOR_CONFIG_FILE";
 const max_line_length = 8192;
 const empty_string = @constCast("");
 const otel_env_var_prefix = "OTEL_";
 
+// Agent paths are empty by default; they are enabled by installing conf.d drop-in files from the
```
jack-berg, line 19: conf.d or injector.d?

.chloggen/config-composable-directory.yaml
```diff
@@ -0,0 +1,37 @@
...
+  The configuration now moves under `/etc/opentelemetry/injector/`, with the main configuration file as
+  `/etc/opentelemetry/injector/otelinject.conf`, merging with files under `/etc/opentelemetry/injector.d`.
+
+  The environment variable configuration file also moves from `/etc/opentelemetry/default_auto_instrumentation_env.conf` to
+  `/etc/opentelemetry/injector/default_env.conf`.
+
+  The default auto-instrumentation agent paths are now empty. Paths are provided by per-language conf.d drop-in files
+  installed by the respective language packages (e.g., `opentelemetry-java-autoinstrumentation` installs `java.conf`).
+  Users building custom container images must now explicitly configure the agent paths via conf.d files, the main
+  configuration file, or environment variables.
```
jack-berg: Can you update the PR description to something like this? This is useful context people will want to see if they follow breadcrumbs back to this PR.

jack-berg (approved):

.chloggen/config-composable-directory.yaml
basti1302: Good point. I took the liberty to edit the PR description accordingly.

basti1302: *non-blocking:* We could maybe improve this a bit to give clear upgrade instructions:

```suggestion
Breaking change: On systems where custom settings have been configured in `/etc/opentelemetry/otelinject.conf`, these need to move to one of the new locations, for example `/etc/opentelemetry/injector/otelinject.conf`.

Breaking change: The in-code defaults for the auto-instrumentation agent paths are now empty. Paths must be provided via external configuration. Users who use the injector binary directly, but do not use the system packages published by this project, and who also have previously relied on the default auto-instrumentation agent paths must now explicitly configure the agent paths. This can be done either via the main configuration file (default: `/etc/opentelemetry/injector/otelinject.conf`), or conf.d-style files in the configuration directory (e.g. files under `/etc/opentelemetry/injector.d`), or via environment variables.
```

We should discuss https://github.com/open-telemetry/opentelemetry-injector/pull/291#discussion_r3084589392 first though

src/config.zig
basti1302: *Blocking:* I think we need to reconsider the directory layout one more time before merging this.

Right now this PR introduces:
- `/etc/opentelemetry/injector/otelinject.conf` -- the main config file
- `/etc/opentelemetry/injector/default_env.conf` -- env vars
- `/etc/opentelemetry/injector.d/*.conf` -- the conf.d dir

So we now have _two directories_ under `/etc/opentelemetry`, one is called `injector` and one is called `injector.d`. Is this idiomatic for a conf.d style approach? I think this could be somewhat confusing.

How about putting everything under the same directory `/etc/opentelemetry/injector`, and then have the conf.d directy also in that directory.

- `/etc/opentelemetry/injector/injector.conf` -- the main config file
- `/etc/opentelemetry/injector/default_env.conf` -- env vars
- `/etc/opentelemetry/injector/conf.d/*.conf` -- conf.d dir

Everything is under `/etc/opentelemetry/injector/`, the main config file is just called `injector.conf`, and the conf.d directory is literally called `conf.d`. If my memory is correct, this is close to what nginx, X11 and others do.

This would allow the following structure:

```
/etc/opentelemetry/injector/
├── injector.conf          ← main configuration file
├── default_env.conf
└── conf.d/                ← drop-in directory
    ├── 00-java.conf
    ├── 10-nodejs.conf
    └── 99-whatever.conf
```

basti1302 (changes requested):

src/config.zig
```diff
@@ -549,6 +555,57 @@ fn readConfigurationFile(arena_allocator: std.mem.Allocator, cfg_file_path: []co
...
+
+    // Collect and sort file names to ensure deterministic ordering
+    var file_names: std.ArrayList([]const u8) = std.ArrayList([]const u8).initCapacity(arena_allocator, 16) catch {
+        print.printDebug("Failed to allocate memory for conf.d file list", .{});
+        return;
+    };
+    var dir_iter = dir.iterate();
+    while (dir_iter.next() catch null) |entry| {
+        if (entry.kind != .file) continue;
+        if (!std.mem.endsWith(u8, entry.name, ".conf")) continue;
```
basti1302, line 618: *blocking:* I think requiring that all files need to end with `.conf` is also not idiomatic. I think we should read _all_ files in this directory, irrespective of the name.

basti1302: Nevermind, turns out I'm wrong and it is indeed the correct idiom.

atoulme:
+1 on Basti’s suggestion

mmanciop:
> So we now have two directories under /etc/opentelemetry, one is called injector and one is called injector.d. Is this idiomatic for a conf.d style approach?

Well, yes. See [`/etc/sudoers` vs `/etd/sudoers.d`](https://superuser.com/questions/869144/why-does-the-system-have-etc-sudoers-d-how-should-i-edit-it). However, tbh both the "sibling" approach like I implement, and the nested one that Basti proposes are both widespread.

I could even see how rename `conf.d` to something like `injected-sdks.d` or so, to better set expectations about what goes in that folder:

```
/etc/opentelemetry/injector/
├── injector.conf          ← main configuration file
├── default_env.conf
└── injected-sdks.d/       ← drop-in directory
    ├── 00-java.conf
    ├── 10-nodejs.conf
    └── 99-whatever.conf
```

basti1302:
> Well, yes. See [`/etc/sudoers` vs `/etd/sudoers.d`](https://superuser.com/questions/869144/why-does-the-system-have-etc-sudoers-d-how-should-i-edit-it). 

But `/etc/sudoers` is the main config _file_ and its sibling `/etd/sudoers.d` the drop-in _directory_, right? This is different from what the PR currently does. I have not come across any wide-spread use of two sibling _directories_ like we have here, `/etc/opentelemetry/injector` and `/etc/opentelemetry/injector.d`, which are both directories.

The current structure would look like this:

```
/etc/opentelemetry/injector/
├── injector.conf 
├── default_env.conf

/etc/opentelemetry/injector.d/
├── 00-java.conf
├── 10-nodejs.conf
└── 99-whatever.conf
```

I think the `/etc/sudoers` plus `/etd/sudoers.d` is not a good fit for us because we have two default files, the main config file and `default_env.conf`.

basti1302:
> ```
> /etc/opentelemetry/injector/
> ├── injector.conf          ← main configuration file
> ├── default_env.conf
> └── injected-sdks.d/       ← drop-in directory
>     ├── 00-java.conf
>     ├── 10-nodejs.conf
>     └── 99-whatever.conf
> ```

That works for me. I would slightly prefer to not be over-specific with the name `injected-sdks.d/` and use a generic name, because we cannot foresee all possible use cases for this directory. But I'm fine with both.

mmanciop:
> > ```
> > /etc/opentelemetry/injector/
> > ├── injector.conf          ← main configuration file
> > ├── default_env.conf
> > └── injected-sdks.d/       ← drop-in directory
> >     ├── 00-java.conf
> >     ├── 10-nodejs.conf
> >     └── 99-whatever.conf
> > ```
> 
> That works for me. I would slightly prefer to not be over-specific with the name `injected-sdks.d/` and use a generic name, because we cannot foresee all possible use cases for this directory. But I'm fine with both.

Alright, then `/etc/opentelemetry/injector/conf.d` it is. We likely do not have enough clarity about the configurations to go as semantic as I proposed.

src/config.zig
mmanciop: Implemented this in 556e623

src/config.zig
mmanciop: We settled for conf.d

.chloggen/config-composable-directory.yaml
mmanciop: Taken a stab at this in e5fd8da

mmanciop (approved):

basti1302 (approved):
👍

PR #312, opened 2026-04-08
Include / exclude matches all, with conditions ANDed and entries ORd

jack-berg:
Fixes a bug where include_paths was ignored when include_with_arguments was not configured. Corrects the semantics so that each non-empty include setting adds a condition that must be satisfied independently (AND across settings, OR within). Extracts allow/deny evaluation into testable functions in patterns_matcher.zig and updates the README accordingly.


grcevski (approved):
Great catch! LGTM!

grcevski:
There seems to be a compile issue.

basti1302 (approved):

atoulme (approved):

PR #313, opened 2026-04-09
Filter processes before eval libc

### LINKED ISSUE #314: Parsing log level produces garbage values sometimes
### Component(s)

_No response_

### What happened?

**Describe the bug**
OTEL_INJECTOR_LOG_LEVEL="debug" is not parsed correctly, it comes out as:

```
[otel-injector] [3141988] unknown value for OTEL_INJECTOR_LOG_LEVEL: "k/exa" -- valid log levels are "debug", "info", "warn", "error", "none".
```

I haven't been able to pin down exactly in in which scenario this happens, but it does happen on my Ubuntu Linux version 6.17.0-1012-oem (buildd@lcy02-amd64-114) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, GNU ld (GNU Binutils for Ubuntu) 2.42)

**Steps to reproduce**
I just used the java injection with no other configuration like this:

```
OTEL_INJECTOR_LOG_LEVEL="debug" LD_PRELOAD=/.../dist/libotelinject_amd64.so JVM_AUTO_INSTRUMENTATION_AGENT_PATH=/.../opentelemetry-javaagent.jar java -jar target/http-client-demo-1.0.0.jar
```

**What did you expect to see?**
Correctly parsing the LOG_LEVEL env variable

**What did you see instead?**
when "debug" is supplied, "k/exa"is reported with a warning message that it's not a valid debug option. I tried "info", "warn", same result.

I found this when testing the overhead of this PR https://github.com/open-telemetry/opentelemetry-injector/pull/313. The changes in the mentioned PR fix this issue. I'm opening this separate issue so we can track it in case someone else encounters it.

### Version

main

## Environment
OS: Linux version 6.17.0-1012-oem (buildd@lcy02-amd64-114) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, GNU ld (GNU Binutils for Ubuntu) 2.42)
Compiler(if manually compiled): zig 0.15.2

### Log output

```shell
[otel-injector] [3141988] unknown value for OTEL_INJECTOR_LOG_LEVEL: "k/exa" -- valid log levels are "debug", "info", "warn", "error", "none".
```

### Additional context

_No response_

### Tip

jack-berg:
Related to #311.

Processes filtered by include/exclude configuration were still triggering libc detection, producing spurious "failed to identify libc" warnings in the injector's stderr output. This PR moves config file reading and allow/deny evaluation ahead of libc detection. To support this reordering, environment variable lookups during config reading are done by reading /proc/self/environ directly rather than via std.posix.getenv, which depends on libc. Filtered processes now exit before libc detection is ever attempted. An integration test verifies that filtered processes produce no injector stderr output.

Fixes https://github.com/open-telemetry/opentelemetry-injector/issues/314

### DIFF
--- a/src/proc_self_environ_parser.zig
+++ b/src/proc_self_environ_parser.zig
@@ -12,15 +12,68 @@ const testing = std.testing;
 
 const proc_self_environ_path = "/proc/self/environ";
 const otel_injector_log_level_env_var_name = "OTEL_INJECTOR_LOG_LEVEL";
-const otel_injector_log_level_environ_prefix = "OTEL_INJECTOR_LOG_LEVEL=";
 pub const otel_injector_disabled_env_var_name = "OTEL_INJECTOR_DISABLED";
-const otel_injector_disabled_env_var_prefix = "OTEL_INJECTOR_DISABLED=";
+
+const max_getenv_entry_length = 8192;
+const max_getenv_buffer_len = max_getenv_entry_length * 2;
+
+/// Function type for reading environment variables, abstracting over reading from
+/// /proc/self/environ (production) vs std.posix.getenv (tests).
+pub const GetenvFn = *const fn (allocator: std.mem.Allocator, name: []const u8) ?[]u8;
+
+/// Looks up an environment variable by reading /proc/self/environ directly,
+/// without depending on libc or std.posix.getenv.
+/// Returns an allocated copy of the value, or null if not found or on error.
+/// Errors are treated the same as a missing variable: if /proc/self/environ is
+/// unreadable, callers that rely on this function (e.g. for OTEL_INJECTOR_CONFIG_FILE)
+/// will silently fall back to their defaults. This is intentional — initFromProcSelfEnviron
+/// runs earlier and will have already surfaced any read failure at a higher log level.
+/// The caller is responsible for freeing the returned slice.
+pub fn getenv(allocator: std.mem.Allocator, name: []const u8) ?[]u8 {
+    return getenvFromFile(allocator, proc_self_environ_path, name) catch null;
+}
+
+/// Wraps std.posix.getenv with the GetenvFn signature for use in tests,
+/// where environment variables are set via setenv()/putenv() rather than
+/// being present in /proc/self/environ.
+pub fn posixGetenv(allocator: std.mem.Allocator, name: []const u8) ?[]u8 {
+    const val = std.posix.getenv(name) orelse return null;
+    return allocator.dupe(u8, val) catch unreachable;
+}
+
+fn getenvFromFile(allocator: std.mem.Allocator, path: []const u8, name: []const u8) !?[]u8 {
+    var environ_file = try std.fs.openFileAbsolute(path, .{});
+    defer environ_file.close();
+    var buf: [max_getenv_buffer_len]u8 = undefined;
+    var reader = environ_file.reader(&buf);
+    while (takeSentinelOrDiscardOverlyLongLine(&reader)) |environ_entry| {
+        if (matchEntry(environ_entry, name)) |value| {
+            return try allocator.dupe(u8, value);
+        }
+    } else |err| switch (err) {
+        error.EndOfStream => {
+            var last_buf: [max_getenv_entry_length]u8 = undefined;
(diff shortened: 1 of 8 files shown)


src/proc_self_environ_parser.zig
```diff
@@ -36,58 +89,40 @@ fn initFromEnvironFile(self_environ_path: []const u8) !void {
...
-            continue;
-        }
-        if (std.mem.startsWith(u8, environ_entry, otel_injector_disabled_env_var_prefix)) {
-            const otel_injector_disabled_env_var_value =
-                environ_entry[otel_injector_disabled_env_var_prefix.len..environ_entry.len];
-            proc_self_environ_values.setOtelInjectorDisabled(parseBooleanValue(otel_injector_disabled_env_var_value));
-            continue;
-        }
-    } else |err| switch (err) {
+    const log_level_value = getenvFromFile(allocator, self_environ_path, otel_injector_log_level_env_var_name) catch |err| switch (err) {
```
grcevski, line 94: My only concern would be the number of time we call `getenvFromFile`. This is going to read the environ repeatedly correct? For a small number of environment variables this is negligible, but it will slow down application startup for startup sensitive programs.

jack-berg: I was thinking about this as well. Some considerations:

- The /proc/self/environ is a virtual file, so traversal is fast
- Can reduce number of traversals by instead doing a single traversal and storing the env vars we care about in a map or array, but this adds complexity / allocation. Also related: the code here introduced a GetenvFn function type so that config reading can avoid using std.posix.getenv. Right now GetenvFn is stateless. If we want to reduce traversal, it needs to carry state. 

I'm open to working to reduce traversal if we think it will be a problem for sensitive programs.

grcevski: I took timing from your new branch and here's the result:
```
[otel-injector] [3135591] getenvFromFile took 4318ns
[otel-injector] [3135591] getenvFromFile took 3957ns
[otel-injector] [3135591] getenvFromFile took 2655ns
[otel-injector] [3135591] getenvFromFile took 3457ns
[otel-injector] [3135591] getenvFromFile took 3406ns
[otel-injector] [3135591] getenvFromFile took 3457ns
[otel-injector] [3135591] getenvFromFile took 3687ns
[otel-injector] [3135591] getenvFromFile took 3767ns
[otel-injector] [3135591] getenvFromFile took 3497ns
[otel-injector] [3135591] getenvFromFile took 3887ns
```

So previously this was only one read of 4us, now it's 10x. I think this is still of the order of 40us, so no big deal IMO. 

At the same time I think you've fixed another bug, when I run `main` with my small simple command:

```
OTEL_INJECTOR_LOG_LEVEL="debug" LD_PRELOAD=/home/nino/work/opentelemetry-injector/dist/libotelinject_amd64.so JVM_AUTO_INSTRUMENTATION_AGENT_PATH=/home/nino/Downloads/opentelemetry-javaagent.jar java -jar target/http-client-demo-1.0.0.jar
```

I get this to print on the stdout:
```
[otel-injector] [3141988] unknown value for OTEL_INJECTOR_LOG_LEVEL: "k/exa" -- valid log levels are "debug", "info", "warn", "error", "none".
```
And obviously setting debug doesn't work.

But with your changes this works.

grcevski (approved):
LGTM! I think the timing increase is negligible and the code is a lot more readable and easier to understand now.

src/proc_self_environ_parser.zig
grcevski: I opened an issue for the bug I found while testing your changes which is in "main". I've linked in this PR with Fixes...

basti1302 (approved):

jack-berg:
Merging will produce merge conflicts with #291 @mmanciop. I'm happy to resolve merge conflicts here or let you resolve them in #291 🙂

mmanciop:
> Merging will produce merge conflicts with #291 @mmanciop. I'm happy to resolve merge conflicts here or let you resolve them in #291 🙂 

I think we agreed in the channel
to first merge this PR, then 291. I can deal with the merge conflicts on the other one @jack-berg

PR #322, opened 2026-04-16
Fix libc detection failure on glibc < 2.34 systems (e.g. RHEL 8) for binaries that do not link libdl.so

jack-berg:
Resolves #311.

On glibc < 2.34 (e.g. RHEL 8 / glibc 2.28), `dlsym` is exported by `libdl.so` rather than `libc.so` itself. Binaries that do not link `libdl.so` (such as `grep`, simple C utilities, and many other system programs) would produce a `failed to identify libc` warning and receive no instrumentation. The fix adds a fallback that looks up `setenv` and `__environ` directly from the ELF symbol table when `dlsym` is not found, which succeeds for `libc.so` on all glibc versions.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -74,6 +74,11 @@ The approach taken by the OpenTelemetry injector is as follows:
       If this lookup is successful, we can use it to actually call the `dlsym` function (without declaring a direct
       dependency on it).
 * Finally, use `dlsym` to find the location of the `setenv` and the `__environ` symbol.
+  On glibc < 2.34 (e.g. RHEL 8, Debian Bullseye), `dlsym` is provided by `libdl.so` rather than `libc.so` itself.
+  If the binary does not link `libdl.so`, `dlsym` will not be found in the libc memory range.
+  In that case, the injector falls back to looking up `setenv` and `__environ` directly from the ELF symbol table of
+  the libc memory range, which works because both symbols are always exported directly by `libc.so` on all glibc
+  versions.
 * Use the `__environ` symbol to read the existing environment variables for the process.
 * Use `setenv` to set or modify the required environment variables (`NODE_OPTIONS`, `JAVA_TOOL_OPTIONS`,
   `OTEL_RESOURCE_ATTRIBUTES` etc.)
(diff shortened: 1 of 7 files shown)


grcevski (approved):
LGTM! I just don't know enough to comment on the removal of CannotFindDlSymSymbol. Previously we purposefully returned that in the event nothing can be found.

src/libc.zig
```diff
@@ -1004,17 +1005,38 @@ fn tryToFindSymbolsInMemoryRange(
         return error.CannotOpenLibc;
     };
 
-    const dlsym_fn =
-        linker.lookup(types.DlSymFn, dlsym_function_name) orelse return error.CannotFindDlSymSymbol;
+    if (linker.lookup(types.DlSymFn, dlsym_function_name)) |dlsym_fn| {
```
mmanciop: I think we need to rename `linker` to `libc` or something similar. When I read it, not having seen this code path in a while, I had a lot of questions in my head. Originally it made sense, because we were looking up symbols in the address space of the linker library (which, in the case of musl, it's the same as libc). But now it is entirely confusing.

mmanciop:
> LGTM! I just don't know enough to comment on the removal of CannotFindDlSymSymbol. Previously we purposefully returned that in the event nothing can be found. 

It's fine to remove it, since we are not anymore giving up when that happens. Now instead, we will give up on finding __environ most likely.

basti1302:
This is awesome stuff! ❤️ 

Two minor comments:

1. If we can lookup the symbols we _actually_ need (setenv, __environ) without dlsym, is there a motivation/reason for keeping both code paths? Would we lose something if we make the "fallback" implemented here the one and only way to get these symbols, and remove the attempt to look up dlsym first? Also cc @mmanciop, since this was your original design. (Disclaimer: I might totally be overlooking something obvious here.)
2. Just as an aside, the "libc but no libdl" issues actually came up already when we Michele and me worked on the initial Zig injector PR, see [DESIGN.md](https://github.com/open-telemetry/opentelemetry-injector/blob/main/DESIGN.md), search for `libdl`:
    > On most modern distributions, the actual libc file (say, libc.so.6) contains getenv, setenv etc. and also dlsym. However, on older distributions (Debian bullseye being one example), dlsym is actually provided by a separate file. Nearly all of libc's functions are provided by libc-2.31.so or similar (which is symlinked as libc.so.6), but that file does not contain the symbols dlsym, dlopen etc. Instead, these symbols are provided by libdl-2.31.so. Most interesting binaries will link both, libc and libdl, but of course binaries can also only link libc, if they do not depend on libdl at all. The best thing the injector could do in this case is to stand down. It will not crash the executable, since there is no linking error. But since it wasn't able to find dlsym, it can also not lookup __environ, hence it has no knowledge of the current environment, hence its getenv override will not be able to return any values for any environment variable. Effectively, this would start the executable with an empty environment, which is not acceptable. This could potentially be worked around by looking up __environ directly without using dlsym, or by falling back to backfilling the environment by reading /proc/self/environ. However, the next issue is even more severe.

It's really great to see this being improved now. I wonder if we need to update DESIGN.md slightly, since the "The best thing the injector could do in this case is to stand down." is no longer true now.

injector-integration-tests/runtimes/no-libdl/app.c
```diff
@@ -0,0 +1,26 @@
...
+// A minimal C program that prints the value of a named environment variable.
+// Deliberately uses only libc (stdio, stdlib) and is compiled without -ldl, so
+// libdl.so does not appear in /proc/self/maps at runtime. This reproduces the
+// scenario seen on RHEL 8 and any other pre-glibc-2.34 system where dlsym lives
+// in libdl.so (not in libc.so), and the binary under test does not link libdl.
+
+#include <stdio.h>
+#include <stdlib.h>
+
+int main(int argc, char *argv[]) {
```
basti1302, line 13: _praise:_ excellent to also have an integration test for this scenario now 👍

jack-berg: And one that I verified fails without this change! 😉

jack-berg:
> If we can lookup the symbols we actually need (setenv, __environ) without dlsym, is there a motivation/reason for keeping both code paths? Would we lose something if we make the "fallback" implemented here the one and only way to get these symbols, and remove the attempt to look up dlsym first? Also cc @mmanciop, since this was your original design. (Disclaimer: I might totally be overlooking something obvious here.)

Caveat that I'm not an expert on this stuff, but based on my research, preferring dlsym over looking up directly in the libraries symbol table is mostly a correctness principle - "let the dynamic linker decide". I believe another LD_PRELOAD shared library could technically change the location.

> It's really great to see this being improved now. I wonder if we need to update DESIGN.md slightly, since the "The best thing the injector could do in this case is to stand down." is no longer true now.

Updated DESIGN.md to reflect to updates:

- Delete the section that calls out debian bullseye and other older distributions which do not contain dlsym as problematic
- Update the # Design section where we talk about using dlsym to resolve the location to mention the fallback direct lookup in the ELF symbol table

src/libc.zig
jack-berg: 👍

mmanciop:
> 1. If we can lookup the symbols we _actually_ need (setenv, __environ) without dlsym, is there a motivation/reason for keeping both code paths?

As I was researching ELF back when, I read about different versions of LibC doing different indirections for symbols in various tables, and that is why I said "guess what, I don't want to debug forever, let's use dlsym instead". With Jack's patch, worst thing that happens is that we fail anyhow.

basti1302:
> > 1. If we can lookup the symbols we _actually_ need (setenv, __environ) without dlsym, is there a motivation/reason for keeping both code paths?
> 
> As I was researching ELF back when, I read about different versions of LibC doing different indirections for symbols in various tables, and that is why I said "guess what, I don't want to debug forever, let's use dlsym instead". With Jack's patch, worst thing that happens is that we fail anyhow.

Sounds good, then I guess we should keep that as the preferred code path.

DESIGN.md
```diff
@@ -208,30 +213,14 @@ read via `getenv` by some runtimes, as explained above)?
...
-   This would lead to the injector reporting `OPENSSL_armcap=null` to OpenSSL, even if it is actually set.
-   Obviously, this is also not acceptable, hence exporting `getenv` is not a viable approach.
+There is an issue that makes this particular strategy prohibitive:
+there can be shared libraries that look up environment variables very early in the startup process, even before the
+injector had a chance to run its initialization code.
+A prominent example is OpenSSL, which is used by many runtimes and applications.
+The OpenSSL code, when run on arm64 CPUS, reads `OPENSSL_armcap` (a capabilities bitmask) before the injector's init
+code runs, that is, before it even had the chance to find the `__environ` pointer and read the process environment.
+This would lead to the injector reporting `OPENSSL_armcap=null` to OpenSSL, even if it is actually set.
+Obviously, this is also not acceptable, hence exporting `getenv` is not a viable approach.
```
basti1302, line 223: Thanks for updating this 👍

basti1302 (approved):
Great stuff!

PR #328, opened 2026-04-29
feat(dotnet): skip injection when profiler env vars already exist

### LINKED ISSUE #270: Do not unconditionally overwrite existing CORECLR_PROFILER and other .NET related environment variables
For other supported runtimes, we check if the associated auto-instrumentation environment variable exists (e.g. NODE_OPTIONS, JAVA_TOOL_OPTIONS, PYTHONPATH). If they already exist, we append (or prepend) "our" part to the existing value.

This is not feasible for .NET, as you can only attach one profiler via CORECLR_PROFILER and the other .NET env vars. The current behavior unconditionally overwrites existing .NET related environment variables, without even checking if they already exist.

The safer option would probably be to check if any of the vars we want to set already exists, and if so, do not set any .NET env vars.

matt-hensley:
Fixes #270

Checks for whether the following env vars are defined:
* CORECLR_ENABLE_PROFILING
* CORECLR_PROFILER
* CORECLR_PROFILER_PATH
* DOTNET_ADDITIONAL_DEPS
* DOTNET_SHARED_STORE
* DOTNET_STARTUP_HOOKS

If any are already set the injector will not overwrite the values.

### DIFF
--- a/.chloggen/dotnet-existing-profiler-env-guard.yaml
+++ b/.chloggen/dotnet-existing-profiler-env-guard.yaml
@@ -1,0 +1,10 @@
+change_type: bug_fix
+component: injector
+note: Do not overwrite existing .NET profiler and startup-hook environment variables when deciding whether to inject .NET auto-instrumentation
+issues: [270]
+subtext: |
+  If the target process already defines `CORECLR_ENABLE_PROFILING`, `CORECLR_PROFILER`,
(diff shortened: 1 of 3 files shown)


mmanciop:
Update: yes, I must have hallucinated it.

~@basti1302 I would have sworn we had this already covered. Did I hallucinate that like a bad LLM?~

mmanciop (approved):

grcevski (approved):
LGTM! Awesome!

basti1302 (approved):
Awesome! Thanks for fixing this 👍

PR #360, opened 2026-06-08
feat(dotnet): skip injection for unsupported runtime versions

matt-hensley:
Attempts to examine the `[application-name].runtimeconfig.json` and read the `tfm` ([target framework moniker](https://learn.microsoft.com/en-us/dotnet/standard/frameworks)) field. The .NET runtime uses this field to determine which API version the runtime should expose the application.

* When the `tfm` value is `net8.0` or higher, injection is allowed to move forward.
* Older `tfm` values like `netcoreapp3.1` will be rejected as eligible for injection.
* If the file or field are missing, the checks fail open.

### DIFF
--- a/src/dotnet.zig
+++ b/src/dotnet.zig
@@ -693,14 +693,14 @@ test "runtimeConfigTargetsModernDotnet: true for net9.0 Microsoft.NETCore.App" {
     try test_util.expectWithMessage(try runtimeConfigTargetsModernDotnet(testing.allocator, content), "runtimeconfig should qualify");
 }
 
-test "runtimeConfigTargetsModernDotnet: true for net10.0 future multi-digit major" {
+test "runtimeConfigTargetsModernDotnet: true for net11.0 future multi-digit major" {
     const content =
         \\{
         \\  "runtimeOptions": {
-        \\    "tfm": "net10.0",
+        \\    "tfm": "net11.0",
         \\    "framework": {
         \\      "name": "Microsoft.NETCore.App",
-        \\      "version": "10.0.0"
+        \\      "version": "11.0.0"
         \\    }
         \\  }
         \\}
--- a/src/dotnet.zig
+++ b/src/dotnet.zig
@@ -244,11 +244,9 @@ fn shouldInjectDotnet(allocator: std.mem.Allocator) bool {
         print.printDebug("Proceeding with the injection of the .NET OpenTelemetry instrumentation. Could not parse {s} safely: {}", .{ metadata_paths.runtimeconfig_path, err });
         return true;
     };
-    if (runtimeconfig_targets_modern_dotnet) |is_supported| {
-        if (!is_supported) {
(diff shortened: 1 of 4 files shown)


src/dotnet.zig
```diff
@@ -352,11 +372,48 @@ fn jsonObjectKeyLooksLikeOpenTelemetryDependency(key: []const u8) bool {
...
+
+fn tfmTargetsModernDotnet(tfm: []const u8) ?bool {
+    if (!std.mem.startsWith(u8, tfm, "net")) return false;
+
+    const rest = tfm[3..];
+
+    // We MUST find a dot to ensure it's a major.minor modern layout (e.g., net8.0)
+    // This immediately filters out legacy dotless TFMs like "net48" or "net472"
+    const dot_index = std.mem.indexOfScalar(u8, rest, '.') orelse return false;
+    if (dot_index == 0) return null;
```
grcevski: I think this makes sense, `dot` shouldn't be the first character we find, but since we've found the field, should we return `false` instead? Same question about the error on parsing the version.

grcevski (commented):
This looks great to me! Thanks @matt-hensley! 

I just had one question/comment to start a discussion on the topic. I don't know enough about the .NET ecosystem to have a concrete opinion if we should allow instrumentation when we see malformed .NET version. I'm only concerned with the case when we've found the `tfm` field.

src/dotnet.zig
matt-hensley: Simplified the checks to remove the indeterminate state after examining the other language checks more

matt-hensley:
> if we should allow instrumentation when we see malformed .NET version. I'm only concerned with the case when we've found the `tfm` field.

Malformed should probably be rejected, as the modern .NET TFM naming scheme is straightforward: `netMAJOR.MINOR` with an optional platform specific suffix preceded with a dash `netMAJOR.MINOR-PLATFORM`. This PR checks against the `netMAJOR` portion.

There are a handful of outliers that are rare / outside the scope of the injector, like the Samsung Tizen .NET runtime.

grcevski (approved):
LGTM!

src/dotnet.zig
```diff
@@ -547,6 +610,119 @@ test "createDotnetMetadataPaths: apphost path produces deps path" {
...
+        \\    "tfm": "net10.0",
+        \\    "framework": {
+        \\      "name": "Microsoft.NETCore.App",
+        \\      "version": "10.0.0"
+        \\    }
+        \\  }
+        \\}
+    ;
+
+    try test_util.expectWithMessage(try runtimeConfigTargetsModernDotnet(testing.allocator, content), "runtimeconfig should qualify");
```
martincostello: Maybe add 11 as the new "future" and update the comment on this one to not be future?

martincostello (approved):

martincostello:
@matt-hensley Did you forget to push a commit?

martincostello (approved):

PR #385, opened 2026-07-09
[binaries] add s390x and ppc64le binaries

atoulme:
Note I don't add tests for those architectures, as we have no runners to run with those.


mmanciop:
I personally do not feel comfortable in distributing binaries we can neither test nor troubleshoot

atoulme:
> I personally do not feel comfortable in distributing binaries we can neither test nor troubleshoot

Working on this, the mainframe SIG is looking into it.

PR #394, opened 2026-07-20
chore(build): derive release bump from chloggen entries

### LINKED ISSUE #380: Simplify release process by selecting `major`/`minor`/`patch`
### Component(s)

_No response_

### Describe the issue you're reporting

The release process is very automated, but there is an error-prone bit by having to type out the entire tag. One could forget the `v` in front, or skipping versions, or creating a tag that is semantically "older", like `v0.8.1` when the current version is `v0.9.0`.

To avoid this issue, should change the type of input to be an enum with `major`, `minor` and `patch` as options, default `minor`.

### Tip

mmanciop:
Replace the free-form `version` input on the prepare-release workflow with an `override` choice input (`auto` / `major` / `minor` / `patch`, default `auto`).

Classification: `breaking` → major, `deprecation` / `new_component` / `enhancement` → minor, `bug_fix` → patch. While the release stream is on 0.y.z, a `breaking` change is bumped as minor per the semver 0.x convention.

Fixes #380

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -2,11 +2,19 @@ name: Automation - Prepare Release
 
 on:
   workflow_dispatch:
-    # Determine the version number that will be assigned to the release.
     inputs:
-      version:
+      override:
+        description: |
+          Override the auto-derived bump. Leave as "auto" to derive from
+          chloggen entries in .chloggen/, or force a specific bump kind.
         required: true
-        description: Release version or release candidate version (like v0.70.0 or v0.70.0-20260109)
+        default: auto
+        type: choice
+        options:
+          - auto
+          - major
+          - minor
+          - patch
 permissions:
   contents: read
 jobs:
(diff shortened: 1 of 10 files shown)


tooling/nextversion/main.go
```diff
@@ -0,0 +1,155 @@
...
+		if err != nil {
+			return BumpNone, err
+		}
+		fmt.Fprintf(os.Stderr, "Bump overridden to: %s\n", kind)
+		return kind, nil
+	}
+
+	changeTypes, err := readChloggenChangeTypes(chloggenDir)
+	if err != nil {
+		return BumpNone, fmt.Errorf("reading chloggen entries from %s: %w", chloggenDir, err)
```
basti1302, line 102: _issue:_ This looks like a release will fail if we ever want to publish a release for changes that come without any chloggen entries (e.g. only chores or PRs with "Skip Changelog"). Maybe that's a non-issue or even desirable, not sure. 

Alternatively we could go with a patch release if there are no chloggen entries.

Use cases for such release might be a release that just bumps dependencies, I think those are currently chore PRs without chloggen entries. OTOH, with the auto-instrumentation agents moving over to packaging, we do not have many dependencies left.

.github/workflows/prepare-release.yml
```diff
@@ -2,11 +2,19 @@ name: Automation - Prepare Release
...
+          chloggen entries in .chloggen/, or force a specific bump kind.
         required: true
-        description: Release version or release candidate version (like v0.70.0 or v0.70.0-20260109)
+        default: auto
+        type: choice
+        options:
+          - auto
+          - major
+          - minor
+          - patch
```
basti1302, line 17: _question:_ Should we include inputs to publish a release candidate/pre-release. AFAICT the current inputs would not allow that. OTOH, we didn't do that in quite a while, so maybe it's not worth it.

basti1302 (approved):

tooling/nextversion/main.go
mmanciop: I am not sure what is the scenario in which we may want to make a release with only chores. Our only release artifact is effectively the .so now.

.github/workflows/prepare-release.yml
mmanciop: Not sure this is warranted, but it might be again with packages moved out? I'll add the mechanics for it.

tooling/nextversion/main.go
basti1302: You are probably right. Let's not do that; we can always add it later in case we ever have the need.

.github/workflows/prepare-release.yml
mmanciop: See c33a89e

PR #408, opened 2026-07-30
fix(elf): guard against empty GNU hash bucket to prevent integer overflow panic

### LINKED ISSUE #399: panic: integer overflow in ElfDynLib.lookupAddress on empty GNU hash bucket, crashes host process (SIGABRT)
### Component(s)

_No response_

### What happened?

**Describe the bug**

When the injector is loaded via `ld.so.preload` on a glibc < 2.34 system (RHEL 8), some host processes abort with `panic: integer overflow` before their own code runs:

```
$ /opt/zextras/common/sbin/clamd --foreground=true --config-file=/opt/zextras/conf/clamd.conf
panic: integer overflow
aborting due to recursive panic
Aborted (core dumped)
```

The panic is raised inside `libotelinject.so` in `ElfDynLib.lookupAddress` (`src/elf.zig`), during the glibc < 2.34 "second pass" of `getLibCInfo` that scans all mapped ELF objects for `dlsym`/`setenv`/`getenv`. Since the injector is preloaded globally, this kills arbitrary system daemons — in our case `clamd`, `freshclam`, and `saslauthd` on a Carbonio mail server, taking down the antivirus service. This defeats the injector's design goal of standing down gracefully instead of crashing the host process.

**Steps to reproduce**

1. RHEL 8 (glibc 2.28, so the glibc < 2.34 second-pass fallback in `getLibCInfo` is active), injector installed with `ld.so.preload` active
2. Start any dynamically linked executable that maps a shared object whose `.gnu.hash` bloom filter false-positives for `setenv` (or `dlsym`/`getenv`) with an empty hash bucket — e.g. ICU 74.2's `libicui18n.so.74.2` as shipped by Carbonio
3. The process aborts before `main()`

**What did you expect to see?**

The injector stands down for symbols/libraries it cannot resolve and lets the host process start uninstrumented.

**What did you see instead?**

SIGABRT of the host process before its own code runs. Core dump backtrace:

```
libotelinject.so
debug.FullPanic.integerOverflow()
elf.ElfDynLib.lookupAddress()
libc.tryToFindSymbolsInMemoryRange()
libc.getLibCInfo()
root.initEnviron()
```

### Version

v0.9.0 (via `opentelemetry-injector-0.0.2-1.x86_64` RPM from opentelemetry-packaging); the affected code is unchanged at current `main`.

## Environment
OS: RHEL 8.10 (glibc 2.28), x86_64
Injector loaded globally via /etc/ld.so.preload
Host processes affected: clamd, freshclam, saslauthd (Carbonio 26.6.1)

### Log output

```shell
$ OTEL_INJECTOR_LOG_LEVEL=debug /opt/zextras/common/sbin/clamd --foreground=true --config-file=/opt/zextras/conf/clamd.conf
...
[otel-injector] [ 202043] attempting dlsym lookup in second pass for /proc/self/maps line: 7f917d90f000-7f917dafa000 r-xp 00000000 fd:00 3948383  /opt/zextras/common/lib/libicuuc.so.74.2
[otel-injector] [ 202043] dlsym lookup in second pass failed for /proc/self/maps line: 7f917d90f000-7f917dafa000 r-xp 00000000 fd:00 3948383  /opt/zextras/common/lib/libicuuc.so.74.2 -- error.CannotFindSetenvSymbol
[otel-injector] [ 202043] attempting dlsym lookup in second pass for /proc/self/maps line: 7f917dd0f000-7f917e028000 r-xp 00000000 fd:00 3948371  /opt/zextras/common/lib/libicui18n.so.74.2
panic: integer overflow
aborting due to recursive panic
Aborted (core dumped)
```

### Additional context

Root cause: in the GNU hash section format, a bucket value of `0` means "no symbols hash to this bucket". glibc's own `dl-lookup` checks for this before computing the chain start. `lookupAddress` in `src/elf.zig` does not:

```zig
const bucket_index = hash % gnu_hash_header.nbuckets;
const chain_index = gnu_hash_section.buckets[bucket_index] - gnu_hash_header.symoffset;
```

If `buckets[bucket_index] == 0` (or any value `< symoffset`), the `u32` subtraction underflows. In a `ReleaseSafe` build this is a checked-arithmetic panic → `abort()`; in an unchecked build it would read out-of-bounds chain memory instead, which is arguably worse.

The path is only reached when the bloom filter false-positives for the looked-up symbol, which is why the crash is host-specific: it depends on the exact set of mapped libraries and their hash layouts. Verified empirically by parsing the `.gnu.hash` section of the crashing library:

```
symbol lookup in /opt/zextras/common/lib/libicui18n.so.74.2
  nbuckets=8209 symoffset=726 bloom_size=1024 bloom_shift=16
  dlsym      hash=0x0f4dc4ae bloom=miss                       -> returns null, OK
  setenv     hash=0x1b85483a bloom=PASS (false positive)
             bucket[0x1b85483a % 8209 = 4405] = 0, symoffset = 726
             0 - 726 as u32 -> underflow -> panic
```

So: second pass scans `libicui18n.so.74.2` → `dlsym` lookup misses the bloom filter (fine) → fallback `setenv` lookup passes the bloom filter as a false positive → empty bucket → underflow → abort. Any process that maps a library with this property crashes. ICU is very widely deployed, and bloom false positives are a statistical certainty across enough libraries — the second-pass scan over *all* mappings makes hitting this likely on real systems.

Suggested fix — treat bucket values below `symoffset` as "symbol not present", mirroring glibc:

```zig
const bucket_index = hash % gnu_hash_header.nbuckets;
const bucket_value = gnu_hash_section.buckets[bucket_index];
if (bucket_value < gnu_hash_header.symoffset) {
    // Empty bucket (0) or malformed table: symbol is not in this object.
    return null;
}
const chain_index = bucket_value - gnu_hash_header.symoffset;
```

Additionally, the chain walk (`while (!at_end_of_chain) : (current_index += 1)`) has no upper bound; a malformed or truncated chain would read past the mapping. Bounding it by the symbol count (or the mapped range) would harden the same function against the unchecked variant of this bug.

More broadly: given the injector's design goal of never crashing the host process, it may be worth catching panics in the init path (or building with a panic handler that logs and stands down) so that any future defect in the ELF-parsing code degrades to "no instrumentation" instead of SIGABRT of arbitrary system daemons.

Workarounds: `OTEL_INJECTOR_DISABLED=true` for affected processes, `exclude_paths` for the affected binaries, or removing the injector package.

### Disclaimer

As I'm not that familiar with Zig and low level components of linux, I've used ai (claude) to debug this issue. Any feedback is welcomed!

### Tip

mmanciop:
Fixes #399.

On glibc < 2.34 systems (RHEL 8, Debian Bullseye, …) the injector's second pass over `/proc/self/maps` calls `ElfDynLib.lookupAddress` on every mapped shared object. The GNU hash lookup did not guard against empty buckets (value 0) or bucket values below `symoffset`:

```zig
// before
const chain_index = gnu_hash_section.buckets[bucket_index] - gnu_hash_header.symoffset;
```

If the bloom filter false-positives for a looked-up symbol and the corresponding bucket is empty, this `u32` subtraction underflows and triggers a ReleaseSafe `panic: integer overflow` that aborts the host process with `SIGABRT` before its own code runs — taking down arbitrary system daemons loaded via `ld.so.preload`.

The fix mirrors glibc's `dl-lookup` behaviour: treat any bucket value below `symoffset` as "symbol not in this object" and return `null`.

```zig
// after
const bucket_value = gnu_hash_section.buckets[bucket_index];
if (bucket_value < gnu_hash_header.symoffset) {
    return null;
}
const chain_index = bucket_value - gnu_hash_header.symoffset;
```

### DIFF
--- a/.chloggen/fix-empty-gnu-hash-bucket-panic.yaml
+++ b/.chloggen/fix-empty-gnu-hash-bucket-panic.yaml
@@ -1,0 +1,16 @@
+change_type: bug_fix
+
+component: injector
+
+note: Fix integer overflow panic when a GNU hash bucket is empty during the second-pass library scan on glibc < 2.34 systems.
+
+issues: [399]
+
+subtext: |
+  On pre-glibc-2.34 systems (e.g. RHEL 8, Debian Bullseye) the injector's second-pass scan of
+  `/proc/self/maps` calls `ElfDynLib.lookupAddress` on every mapped shared object. If the GNU
+  hash bloom filter false-positives for a looked-up symbol and the corresponding bucket value is
+  zero (empty bucket) or otherwise less than `symoffset`, the `u32` subtraction underflows and
+  triggers a ReleaseSafe `panic: integer overflow` that aborts the host process with `SIGABRT`
+  before its own code runs. The fix mirrors glibc's `dl-lookup` behaviour: treat any bucket
+  value below `symoffset` as "symbol not present" and return `null`.
--- a/injector-integration-tests/apps/empty-gnu-hash-bucket/Dockerfile
+++ b/injector-integration-tests/apps/empty-gnu-hash-bucket/Dockerfile
@@ -1,0 +1,62 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Regression test for issue #399:
+# "panic: integer overflow in ElfDynLib.lookupAddress on empty GNU hash bucket"
+#
+# debian:bullseye-slim ships glibc 2.31. On glibc < 2.34, /proc/self/maps shows
+# "libc-2.31.so" not "libc.so.6", so the injector's first pass finds nothing and
+# the second pass scans ALL mapped .so files.
+#
+# A crafted library (trigger.so) has a .gnu.hash section where the bloom filter
+# passes (false positive) for "setenv" but the corresponding bucket value is 0
+# (empty bucket). Without the fix, this causes:
+#   chain_index = 0 - symoffset(1) → u32 underflow → panic: integer overflow → SIGABRT
+#
+# premapper.so maps trigger.so at a fixed low address (0x1000000) before the
+# injector's constructor runs, so the second-pass scan hits it before libc.
+# Because premapper.so is listed second in LD_PRELOAD, its constructor runs first
+# (the dynamic linker initialises later entries first when there are no inter-
+# library dependencies), ensuring trigger.so is already mapped when the injector's
+# initEnviron is called.
+ARG base_image_run="debian:bullseye-slim@sha256:0083feb8da4f624e3a0245e7752af2517d4b81d8b8db50c725644672a132a31b"
+FROM ${base_image_run}
+
+ARG injector_binary
+RUN if [ -z "$injector_binary" ]; then \
+      echo "Error: build argument injector_binary is required but not set."; \
+      exit 1; \
+    fi
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends gcc libc6-dev python3 \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /etc/opentelemetry/injector/conf.d && chmod -R 777 /etc/opentelemetry
+
+# Generate the crafted trigger.so with the malformed GNU hash section.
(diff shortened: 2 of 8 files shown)


atoulme (approved):
wow that's a nice fix. Thank you!

Issue #406, opened 2026-07-30
Make configurable the versions of runtimes the Injector will inject


mmanciop:
### Is your feature request related to a problem? Please describe.

For .NET, we hard-coded the injection to .NET 8 and above. No similar safeguards exist for other runtimes, although we know that Python needs them (2.7x, 3.[<10]).

### Describe the solution you'd like

Make that configurable, with System Packaging and OpenTelemetry Operator being responsible to configure the versions of runtimes that should be injected to match the support matrix of the injected SDK.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

riccardomenegazzo:
I'd like to help with the next part of this.

I looked at the current state before starting an implementation. #409 already made the .NET minimum version configurable. For Python, there is an important wrinkle: the current `opentelemetry-packaging` Python bundle already has a version gate in `sitecustomize.py` (currently Python >= 3.10) and self-deactivates on unsupported interpreters.

Because the injector deliberately applies environment changes broadly and generally fails open when it cannot identify the target runtime, duplicating that policy in `python.zig` would require reliable runtime/version detection before Python initializes and could change inheritance behavior.

My preference would be to keep the injector generic: make compatibility bounds configurable, let packaging/operator own the supported-version policy, and only stand down when an unsupported runtime version can be identified with confidence; unknown versions should preserve today's fail-open behavior.

If that matches the intended direction for #406, I can take the Python part first and include unit/integration coverage plus documentation.

PR #409, opened 2026-07-30
feat(dotnet): make the minimum .NET major version required for injection configurable

mmanciop:
## Description

The injector inspects the `*.runtimeconfig.json` file of the target application and skips the injection of the .NET auto-instrumentation if the application does not target a supported .NET version. Previously, the minimum supported .NET major version was hardcoded to `8`, matching the .NET versions supported by the upstream OpenTelemetry .NET auto-instrumentation.

Distributions of the .NET auto-instrumentation that support older .NET versions need to be able to lower this threshold. This change makes it configurable via:

- Configuration file key: `dotnet_auto_instrumentation_minimum_dotnet_major_version`
- Environment variable: `DOTNET_AUTO_INSTRUMENTATION_MINIMUM_DOTNET_MAJOR_VERSION`

The default remains `8`, so there is no change in behavior for existing users.

## Changes

- `src/config.zig`: adds parsing and validation for the new `dotnet_auto_instrumentation_minimum_dotnet_major_version` config key and corresponding env var
- `src/dotnet.zig`: threads the configurable minimum version through the .NET version check logic, replacing the hardcoded `8`
- `README.md`: documents the new configuration key and env var
- `unit-test-assets/config/all_values.conf`: adds the new key to the exhaustive test config file
- `.chloggen/dotnet-minimum-major-version-configurable.yaml`: changelog entry

First PR related with #406

### DIFF
--- a/src/ruby.zig
+++ b/src/ruby.zig
@@ -395,6 +395,7 @@ fn testConfiguration(path_prefix: []const u8, disabled: bool) config.InjectorCon
         .exclude_paths = &.{},
         .include_args = &.{},
         .exclude_args = &.{},
+        .dotnet_auto_instrumentation_minimum_dotnet_major_version = config.default_dotnet_auto_instrumentation_minimum_dotnet_major_version,
         .dotnet_instrumentation_disabled = false,
         .jvm_instrumentation_disabled = false,
         .nodejs_instrumentation_disabled = false,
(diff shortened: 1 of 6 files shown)


atoulme (approved):

src/ruby.zig
```diff
@@ -395,6 +395,7 @@ fn testConfiguration(path_prefix: []const u8, disabled: bool) config.InjectorCon
         .exclude_paths = &.{},
         .include_args = &.{},
         .exclude_args = &.{},
+        .dotnet_auto_instrumentation_minimum_dotnet_major_version = config.default_dotnet_auto_instrumentation_minimum_dotnet_major_version,
```
atoulme, line 398: this is probably out of place?

mmanciop: Oh yes, I did not notice the comment before merging, will open follow up

PR #412, opened 2026-07-31
feat: remove architecture gate from .NET injection

### LINKED ISSUE #411: Remove architecture gating for .NET
As discussed in the SIG on Thursday, July 30th, we should remove from the injector the opinionation in terms of under which conditions will the injector inject, and delegate that to system packages / OpenTelemetry Operator and other downstream adopters. Case in point, the hardcoded [refusal to inject .NET](https://github.com/open-telemetry/opentelemetry-injector/blob/fd3e1b0313386ea04eb30f475dcd38df0aedd207/src/dotnet.zig#L788) on anything that is not `arm64`/`amd64`.

mmanciop:
Fixes #411.

The injector previously hard-blocked .NET injection on any architecture that wasn't `x86_64` or `aarch64`, returning `error.UnsupportedCpuArchitecture` and refusing to construct SDK paths at all.

### DIFF
--- a/src/dotnet.zig
+++ b/src/dotnet.zig
@@ -45,7 +45,6 @@ pub const CachedDotnetValues = struct {
 
 const DotnetError = error{
     UnknownLibCFlavor,
-    UnsupportedCpuArchitecture,
     OutOfMemory,
 };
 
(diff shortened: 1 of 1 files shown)


mmanciop:
Closing to reopen from the Dash0 fork (EasyCLA requirement).

PR #428, opened 2026-09-08
test(integration): fix tests based on Debian Bullseye

basti1302:
Debian Bullseye fell out of LTS support end of August 2026. The container images for the integration tests relying on this version can no longer be built:
  Release file for
  http://deb.debian.org/debian-security/dists/bullseye-security/InRelease
  is expired

We still need a distro with a fairly old glibc though, since the test cases are specifically for a regression that only happens in older glibc where the main libc so file does not include dlsym.

This replaces debian:bullseye-slim with a Red Hat UBI 8 base image, which ships glibc 2.28. UBI 8 is supported until 2029.

I also verified that the tests are still valid by temporarily removing their respective fixes, and check that they actually fail.

Additionally, this improves the test script's output, to make it easier to locate the actual failures in the fairly large output of the full test run.

- After running all tests of a test set, list the failed test cases.
- After running all test sets for a libc flavor, list the test sets with failed test cases.
- List the failed test sets in the final summary as well.

### DIFF
--- a/injector-integration-tests/apps/empty-gnu-hash-bucket/Dockerfile
+++ b/injector-integration-tests/apps/empty-gnu-hash-bucket/Dockerfile
@@ -4,9 +4,14 @@
 # Regression test for issue #399:
 # "panic: integer overflow in ElfDynLib.lookupAddress on empty GNU hash bucket"
 #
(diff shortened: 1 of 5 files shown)


jack-berg (approved):

grcevski (approved):
LGTM!

mmanciop (approved):

PR #431, opened 2026-09-11
ci: create version bump PR when a new Zig version is available

basti1302:
The new workflow creates a PR with one simple commit which bumps `ZIG_VERSION` in `zig-version` and `.minimum_zig_version` in `build.zig.zon`.

The automatically created commit does not migrate the code base to the new Zig version, this can hardly be automated. The automatic creation of the pull request is mainly meant as a signal that a new Zig release is available. Its initial CI run will most likely fail, which is expected.

Additionally, it adds a new "make lint" step that verifies that the two Zig versions agree. This was previously only a comment, now it is enforced in CI.

### DIFF
--- a/.github/workflows/scripts/update-zig.sh
+++ b/.github/workflows/scripts/update-zig.sh
@@ -1,0 +1,204 @@
+#!/usr/bin/env bash
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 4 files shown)


basti1302:
Note: Testing GH action workflows is cumbersome as long as a workflow with the same name has not landed on `main` yet (you cannot trigger the workflow from the branch without that, even it has workflow_dispatch). I'll follow by testing the workflow after this PR has landed, potential fixes will then need to land in a follow-up PR.

mmanciop (approved):

PR #433, opened 2026-09-11
tmp: deliberately downgrade Zig version to test the upgrade workflow

basti1302:
Do not merge this.


basti1302:
This is probably moot, see #434.

PR #434, opened 2026-09-11
create a renovate rule to update zig

thompson-tomo:


grcevski:
I'm just wondering here... when this happens to trigger, it would definitely not be a PR that we can merge. Usually zig upgrades require source changes. Is the idea to just catch the new zig release so we don't forget?

thompson-tomo:
Correct, it is part of the reason it would be created as a draft. In effect it replaces the custom script/workflow introduced in https://github.com/open-telemetry/opentelemetry-injector/pull/431 with config for the tooling already in use.

basti1302:
> Correct, it is part of the reason it would be created as a draft. In effect it replaces the custom script/workflow introduced in #431 with config for the tooling already in use.

One very small downside with the renovate approach is that there is no place to make it obvious that the intention is not to create a PR that is ready to be merged, but only serve as a signal that a new Zig version is available. The PR body and commit comment cannot be customized (I think?), and you cannot add comments to renovate.json.

Maybe we could add a sentence or so in CONTRIBUTING.md to explain that?

thompson-tomo:
> One very small downside with the renovate approach is that there is no place to make it obvious that the intention is not to create a PR that is ready to be merged, but only serve as a signal that a new Zig version is available. The PR body and commit comment cannot be customized (I think?), and you cannot add comments to renovate.json.

The pr body & title can in fact be customised and usually the draft is an indicator that it is not ready to merge.

In relation to comments, rules etc have a description field which is freetext and serves the same purpose as coments

basti1302:
> The pr body & title can in fact be customised and usually the draft is an indicator that it is not ready to merge.
> In relation to comments, rules etc have a description field which is freetext and serves the same purpose as coments

I didn't know! Even better then. Maybe we can use both to make the intent clearer, roughly similar to what the script did.

thompson-tomo:
Sure will utilise both when i fix the datasource.

mmanciop (approved):
LGTM about the direction, great contribution @thompson-tomo :-)

renovate.json
```diff
@@ -30,6 +31,39 @@
...
+      ],
+      "datasourceTemplate": "custom.zig",
+      "depNameTemplate": "zig",
+      "depTypeTemplate": "language",
+      "versioningTemplate": "semver-coerced"
+    }
+  ],
+  "customDatasources": {
+    "zig": {
+      "defaultRegistryUrlTemplate": "https://ziglang.org/download/index.json",
```
jamietanna, line 56: We'd also welcome this being added upstream as an official datasource

PR #435, opened 2026-09-21
feat: sign release artifacts with Sigstore

### LINKED ISSUE #350: Add signing to releases
### Component(s)

_No response_

Our .so files are not signed right now and uploaded to github releases.

We should sign the artifacts we push to github as part of releases.

_No response_

### Additional context

_No response_

### Tip

riccardomenegazzo:
## What this changes

Issue #350 asks for the raw injector release artifacts to be signed.

This adds a separate signing step between the existing build and release jobs. On a release tag, the finalized `libotelinject_*.so` files are signed keylessly with Sigstore Cosign, verified, and only then passed to the release job.

The signing and verification logic lives in two small scripts that are used both by regular CI and by the release workflow. This keeps the release path testable without granting OIDC permissions to pull-request jobs.

## Release flow

For a release tag, the workflow now:

1. builds the injector artifacts as before;
2. downloads the finished `.so` files into the signing job;
3. signs each artifact with `cosign sign-blob` and writes a `.sigstore.json` bundle next to it;
4. verifies every bundle against the exact repository, release workflow and tag;
5. uploads the verified artifact set for the existing release job to publish.

If an artifact is missing, a bundle is missing, or verification fails, the release stops.

The release signing job is the only job granted `id-token: write`. The publishing job keeps only `contents: write`.

## Regular CI coverage

Outside release tags, CI downloads the same built artifacts and runs the same signing and verification scripts.

Regular CI generates an ephemeral local Cosign key pair and an empty Cosign signing configuration, so the test path does not use Fulcio, Rekor, a TSA, or GitHub OIDC. The artifacts are signed with that temporary key and the generated bundles are verified with the corresponding public key.

This exercises artifact discovery, bundle creation, the Cosign invocation, failure handling, and bundle verification on normal CI runs while keeping the production release identity and OIDC permissions confined to trusted release events.

## User-facing documentation

The README includes a copy-pasteable verification example that checks both the artifact bytes and the expected GitHub Actions identity. It also makes clear that releases created before signing was introduced will not have Sigstore bundles.

## Validation

The change is rebased onto the current `main`. The workflow and verification model follow the current OpenTelemetry Security SIG recommendations for downloadable release artifacts.

The exact keyless/OIDC release identity can only be exercised on a real release event in the canonical repository; the shared signing and verification code is exercised by regular CI with the ephemeral test key described above.

Closes #350.


atoulme:
Looks like a good idea. Would it be possible to have that done on regular CI as well, so we exercise this code outside of releases?

riccardomenegazzo:
Yes, good point. I moved the signing and verification logic into shared scripts and added a regular CI job that runs them against the same built `.so` artifacts.

For normal CI, the job generates an ephemeral local Cosign key, signs the artifacts without uploading the test signatures to the transparency log, and verifies the bundles with the corresponding public key. Release tags still use keyless GitHub OIDC and verify against the exact workflow/tag identity.

That way the artifact discovery, bundle creation, Cosign invocation, failure handling, and verification path are exercised outside releases, without granting `id-token: write` to pull-request jobs.

Thanks for pointing this out.

atoulme (approved):

riccardomenegazzo:
The new CI job caught a Cosign v3-specific issue on its first run: `--tlog-upload=false` is no longer accepted with the v3 signing configuration path.

I updated the test setup to use an empty Cosign signing config instead. That keeps the CI signing path offline from Fulcio/Rekor/TSA while still exercising the same signing and verification scripts used by releases. Verification with the ephemeral public key is offline as well.

This is exactly the kind of version-specific behavior I was hoping the regular CI coverage would catch.

.github/workflows/build.yml
```diff
@@ -47,10 +47,88 @@ jobs:
           path: ./dist/*.so
           if-no-files-found: error
 
-  publish-stable:
+  test-release-signing:
```
mmanciop, line 50: Why splitting `test-release-signing` and `sign-release-artifacts`? I assume it is for the `write` permissions being excluded for testing, but I am not sure we need that split. Couldn't we instead do something like:
> build → sign keylessly in the release job → verify immediately as a self-check → publish
I understand ^ would be closer to the examples of https://github.com/sigstore/cosign-installer

mmanciop (changes requested):

.github/workflows/build.yml
riccardomenegazzo: Yes, I think that's cleaner.
The separate `test-release-signing` job was added to exercise the signing/verification path in regular CI, following the earlier review feedback, while keeping OIDC out of pull request jobs.
But I agree that the release itself doesn't need a separate signing job. I'll keep the regular CI coverage and move the keyless signing and immediate verification directly into the release/publish job:

`build → sign keylessly → verify → publish`

That also removes the extra signed-artifact upload/download handoff, while the shared scripts are still exercised by both regular CI and the real release path.

Thanks @mmanciop this is simpler!

